Issue #2, opened 2020-10-09
[UI core library] Link


aigoncharov:
https://zeroheight.com/11d0e6dac/p/82559e-link

1. Create a component with API (match new `Tooltip` component)
    1. content (as children)
    2. type

PR #40, opened 2020-10-20
Update Link component

danad02:
Linked to #2


danad02:
verify

packages/ui/src/Link.tsx
```diff
@@ -40,8 +41,8 @@ type LinkProps = IconProps & {
  * - Link can be used as a stand-alone component with right chevron icon.
  */
 export const Link: FC<LinkProps> = ({
-  size = 'regular',
-  Icon,
+  size = 'normal',
+  icon,
```
jankoritak, line 45: Shouldn't we mirror Figma, as for naming? I would personally revert to the previous naming.

jankoritak (approved):
LGTM, leaving a comment regarding naming changes.

packages/ui/src/Link.tsx
danad02: I used `normal` and `small` because we have it everywhere - icon sizes, typography, ...

PR #47, opened 2020-10-22
Enhance mountAndCheckA11Y

aigoncharov:


danad02:
@aigoncharov I just merged Link. Can you update it too?

aigoncharov:
@danad02 it is for the `feat/tooltip` branch. We'll have to update it there if we decide to go with this approach.

danad02 (approved):

packages/test-helpers/src/enzyme.ts
```diff
@@ -14,16 +15,37 @@ export const axeDefaultOptions: AxeOptions = {
...
 }
 
+export interface MountOptions {
+  mountOptions?: MountRendererProps
+  axeOptions?: AxeOptions
+  act?: 'sync' | 'async'
+}
+
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export const mountAndCheckA11Y = async <P>(
```
jvorcak, line 25: @aigoncharov Maybe it'd help if we add some documentation here? I've never used your helpers before and I have to study the code in order to understand what it does.

jvorcak (changes requested):

packages/test-helpers/src/enzyme.ts
aigoncharov: @jvorcak do you think it would be OK to address it later, once we have released the first draft of the `ui`? I have created a separate task for it
https://github.com/hazelcast/frontend-shared/issues/51

jvorcak (approved):

packages/test-helpers/src/enzyme.ts
jvorcak: @aigoncharov sure!

PR #49, opened 2020-10-23
Feature/alert

### LINKED ISSUE #45: [UI core library] Alert
Originates at https://github.com/hazelcast/frontend-shared/issues/4

https://zeroheight.com/11d0e6dac/p/62d703-alert

Create an Alert component with API:
type
icon
content (as children)

jankoritak:
Closes #45 

TODO:
- [x] Button action type
- [x] Weird actions alignment


packages/ui/__stories__/Alert.stories.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
```
aigoncharov, line 1: Great set of stories!

packages/ui/__stories__/Icon.stories.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
```
aigoncharov, line 1: Thank you for doing this!

packages/ui/src/Alert.tsx
```diff
@@ -0,0 +1,122 @@
...
+export type AlertActionButton = {
+  text: string
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void
+} & AccessibleActionButtonIconProps
+
+export type AlertActionLink = {
+  text: string
+} & PartialRequired<AnchorHTMLAttributes<HTMLAnchorElement>, 'href'>
+
+export type AlertAction = AlertActionButton | AlertActionLink
```
aigoncharov, line 34: This is what I call well-thought types! Good job!

packages/ui/src/Toast.module.scss
```diff
@@ -8,9 +8,12 @@
...
   padding: c.$grid * 1.5 c.$grid * 2.5;
 
+  .icon {
+    margin-right: c.$grid * 2.5;
+  }
+
   .content {
     word-break: break-all;
-    margin-left: c.$grid * 2.5;
   }
```
aigoncharov, line 17: Is there any difference?

packages/ui/src/Button.module.scss
```diff
@@ -128,107 +122,9 @@
...
-
-      // TODO: Colours not specified
-      &:disabled {
-        background-color: pink;
-      }
-    }
-  */
+  // Kind === transparent
+  &.transparent {
+    color: c.$colorText;
```
aigoncharov, line 127: Seems like we do not need `color: c.$colorNeutralWhite;` for the `.button` class. If we removed it we could use here `color: inherit` and let the parent container define the color of the button text. It would allow this button to have proper visuals in any parent context. WDYT?

packages/ui/src/Button.tsx
```diff
@@ -1,83 +1,58 @@
...
 
 // Left icon is always present with proper aria-label attribute
 type AccessibleIconLeftProps =
   | {
-      IconLeft: Icon
+      iconLeft: IconProps['icon']
       iconLeftAriaLabel: string
+      iconLeftSize?: IconProps['size']
+      iconLeftColor?: IconProps['color']
+      iconLeftClassName?: string
```
aigoncharov, line 18: nit: Do you think we have a use-case for these extra props right now? Our buttons at this point are of fixed height and colors. So the icons are also of fixed height and colors.

packages/ui/src/Button.tsx
```diff
@@ -98,52 +73,55 @@ export type ButtonProps = ButtonKindPrimarySecondaryProps &
...
+          color={iconLeftColor}
+        />
       )}
-      <TruncatedText text={children.toUpperCase()} />
-      {IconRight && (
+      <TruncatedText text={capitalize ? children.toUpperCase() : children} />
+      {iconRight && iconRightAriaLabel && (
         // Icon colour & size are defined in SCSS
-        <IconRight aria-label={iconRightAriaLabel} data-test="button-icon-right" className={cn(styles.icon, styles.iconRight)} />
+        <Icon
```
aigoncharov, line 117: Good thinking on using the `Icon` component!

packages/ui/src/Alert.module.scss
```diff
@@ -0,0 +1,153 @@
...
+  /* Theme - Success */
+  &.success {
+    color: c.$colorNeutralWhite;
+    background-color: c.$colorSuccess;
+
+    .icon {
+      color: c.$colorNeutralWhite;
+    }
+
+    .closeIcon {
```
aigoncharov, line 50: If we went along with https://github.com/hazelcast/frontend-shared/pull/49/files#r511764453 we could probably remove this class at all

packages/ui/src/Alert.module.scss
```diff
@@ -0,0 +1,153 @@
...
+
+    .title {
+      margin-right: c.$grid * 8.5;
+
+      @include c.typographyHighlightFont;
+      @include c.typographyBodyNormal;
+      @include h.overflowEllipsis;
+    }
+
+    .close {
```
aigoncharov, line 29: The parent container here is flex. Why do we need to make this element absolute? Wouldn't flex parent position it correctly if we added `flex: 1 0 auto;` to `title`?

packages/ui/src/Alert.tsx
```diff
@@ -0,0 +1,122 @@
...
+        [styles.warning]: type === 'warning',
+        [styles.critical]: type === 'critical',
+      })}
+    >
+      <div className={styles.header}>
+        <Icon data-test="alert-icon" ariaLabel={ariaLabel} icon={icon} className={styles.icon} />
+        <div data-test="alert-title" className={styles.title}>
+          {title}
+        </div>
+        <IconButton
```
aigoncharov, line 65: I am wondering if we should add `transparent` type here as well. We also need to think if `#FAFAFA` background works as the hover effect. @pawelklasa WDYT?
The current design of Icon Button (https://www.figma.com/file/8mVm6LTbp2Z0RaWWjTZoft/%F0%9F%90%9DHIVE---Hazelcast-Design-System?node-id=3352%3A8) does not play nice with different background colors.
We would like to use it for the `X` button for our alerts https://www.figma.com/file/8mVm6LTbp2Z0RaWWjTZoft/%F0%9F%90%9DHIVE---Hazelcast-Design-System?node-id=2341%3A111, but the `#FAFAFA` hover does not fit there. Should we do something else for the hover effect? Maybe, we could change the color instead? E.g. for the hover. of the `neutral-white` Icon Button we could make it `neutral0lighter`, for the `colot-text` we could do `secondary-text`. Alternatively, we could darker the background. WDYT?

packages/ui/src/Button.module.scss
jankoritak: Sure, we can do that!

packages/ui/src/Toast.module.scss
jankoritak: None. I honestly don't remember the reasoning behind this. Can roll back, if you prefer to include it in `.content`.

packages/ui/src/Button.tsx
jankoritak: These are defined because we need to tweak the spacing a bit for Alerts actions: https://github.com/hazelcast/frontend-shared/pull/49/files#diff-674337aa6f682273f1632c6b0625a49e0606aa8f8e06facd4ca62a475ed76231R86.

packages/ui/src/Alert.module.scss
jankoritak: Good thinking!

packages/ui/src/Alert.module.scss
jankoritak: Can't get the link working properly. Could you post a snippet of what do you mean?

packages/ui/src/Alert.tsx
jankoritak: I would lean towards changing the `IconButton` hover/rollover. If we can find an accessible alternative, to this `background-color` toggle, great! If not, we can surely add `transparent` `IconButton` type. This is makes sense to me, as we already implement it for regular button, right?

I'm a. bit lost in the last couple of sentences, though.

> Maybe, we could change the color instead? E.g. for the hover. of the neutral-white Icon Button we could make it neutral0lighter, for the colot-text we could do secondary-text. Alternatively, we could darker the background. WDYT?

`IconButton` does not have text, right?

aigoncharov: @jankoritak I meant that for the hover effect we could simply darken the background or darken the icon color.

packages/ui/src/Toast.module.scss
aigoncharov: I asked just out of curiosity. No need to rollback :)

packages/ui/src/Alert.module.scss
aigoncharov: It was linking to the idea of using `color: inherit` for the transparent button.

packages/ui/src/Button.tsx
aigoncharov: Oh, I see. We not only decreased the margin there, but also decreased the font size of the button. Do you think it makes sense to add a `size` prop to the `Button` with possible values `'small' | 'normal'`. For the `small` `Button` we could decrease the margins, change the icon sizes and remove capitalizations. In other words, kind of encapsulate these styling changes. WDYT?

aigoncharov: I think I am basically building on the previous idea we discussed during the initial `Button` implementation. Instead of having lots of props with lots of freedom, we should try created several more or less rigid presets. WDYT?

packages/ui/styles/constants/_typography.scss
```diff
@@ -67,6 +71,13 @@ $lineHeightBodySmall: normal !default;
   letter-spacing: $letterSpacingButton;
   line-height: $lineHeightButton;
 }
+@mixin typographyBodyHighlight {
```
aigoncharov, line 74: I would suggest to parameterize  `typographyBodyNormal`
```scss
@mixin typographyBodyNormal($weight: 'regular') {
  font-family: $fontFamilyText;
  font-size: $fontSizeBodyNormal;
  letter-spacing: $letterSpacingBodyNormal;
  line-height: $lineHeightBodyNormal;

  @if $weight == 'regular' {
    font-weight: $fontWeightTextRegular; 
  }
  @if $weight == 'bold' {
    font-weight: $fontWeightTextBold; 
  }
  @if $weight == 'light' {
    font-weight: $fontWeightTextLight; 
  }
}	}
```

aigoncharov: In fact, we could go even further
```scss
@mixin typographyBodyWeight($weight: 'regular') {
  @if $weight == 'regular' {
    font-weight: $fontWeightTextRegular; 
  }
  @if $weight == 'bold' {
    font-weight: $fontWeightTextBold; 
  }
  @if $weight == 'light' {
    font-weight: $fontWeightTextLight; 
  }
}

@mixin typographyBodyNormal($weight: 'regular') {
  font-family: $fontFamilyText;
  font-size: $fontSizeBodyNormal;
  letter-spacing: $letterSpacingBodyNormal;
  line-height: $lineHeightBodyNormal;

  @include typographyBodyWeight($weight);
}	}
```

packages/ui/styles/constants/_typography.scss
```diff
@@ -39,6 +39,10 @@ $lineHeightBodySmall: normal !default;
   @return '';
 }
 
+/* @mixin typographyHighlightFont {
+  font-family: $fontFamilyOpenSans, sans-serif;
+  font-weight: $fontWeightBold;
+} */
```
aigoncharov, line 45:
```suggestion
```

packages/ui/src/Alert.tsx
jankoritak: I see now. Let's wait for @pawelklasa's POV.

packages/ui/styles/constants/_typography.scss
jankoritak: Great idea!

packages/ui/src/Button.tsx
jankoritak: Conceptually, that makes sense to me, however are we sure that we'll ever use `small` button in any case expect from this one? I would defer this if we have another use-case.

aigoncharov: Fair enough. Let's not add potentially semi-dead code.

aigoncharov (approved):
LGTM!
I think we should not wait for the transparent IconButton design from Pawel and merge this one in. Once we have the design we will refactor.
I would suggest to address https://github.com/hazelcast/frontend-shared/pull/49/files#diff-4a2ce829e49e0bb7a1b6dfdb6fb013be8ed6728646e50bd2707e5e43fd05fab0R46 before merging.
Could you also generate the reference screenshots for the news stories?

packages/ui/src/Alert.module.scss
jankoritak: Was able to get rid of the class in the end. Addressed in: https://github.com/hazelcast/frontend-shared/pull/49/commits/87c15d54aaccef3c99a0047b1eae2ec54c719544

PR #52, opened 2020-10-28
Help has an optional padding property

jvorcak:
Problem statement: We need to have a flexibility to use this `Help` component without a padding e.g. on a newline.

In the same time, it should work OTTB with a naturally looking padding, for example in text without a need to explicitly use a wrapper every time. This approach is used by other design systems as well from what I had a look.

Usage:
```
 <p>
    Lorem Ipsum is simply dummy text of the <span id="test">printing</span>
    <Help parentId="test" helperText="Lorem Ipsum is simply dummy text of the printing and typesetting industry." />
    and typesetting industry.
  </p>
```

or 
```
{helperText && <Help parentId={idRef.current} helperText={helperText} padding="none" />}
```


aigoncharov (approved):
LGTM!

PR #54, opened 2020-10-29
Feature/radio button

jvorcak:
The most common usage will be 

```
type Values = {
      tosApproved: boolean
      name: string
    }

    const onSubmit = jest.fn()

    const TestForm = () => (
      <Formik<Values>
        initialValues={{
          tosApproved: false,
          name: 'Yoda',
        }}
        onSubmit={onSubmit}
      >
        <Form>
          <RadioGroupFieldFormik<Values> name="name">
            <RadioFieldFormik value="aragorn" helperText="The king" label={'Aragorn'} />
            <RadioFieldFormik value="gandalf" helperText="The wizard" label={'Gandalf'} />
          </RadioGroupFieldFormik>
        </Form>
      </Formik>
    )
```


packages/ui/src/Checkbox.module.scss
```diff
@@ -23,8 +23,8 @@
 
   // hide native checkbox
   input {
-    height: 0;
-    width: 0;
+    height: 1px;
```
jvorcak, line 26: I've learned that some older browsers had problems with reading this if height/width is 0px. It might not be a problem today, but it should be safer.

packages/ui/__tests__/RadioGroupFormik.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      wrapper.find('form').simulate('submit')
+    })
+
+    expect(onSubmit).toBeCalledTimes(1)
+    expect(onSubmit).toBeCalledWith(
+      {
+        tosApproved: false,
+        name: 'aragorn',
+      },
+      expect.anything(),
```
jankoritak, line 49: Cool utility, was not aware of this!

packages/ui/__tests__/Radio.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Radio } from '../src/Radio'
+
+jest.mock('uuid')
+
+const uuidMock = uuid as jest.Mock<ReturnType<typeof uuid>>
+
+describe('Radio', () => {
+  beforeEach(() => {
+    uuidMock.mockImplementation(() => 'uuidtest')
+  })
```
jankoritak, line 14: Nit: We can define the `mockImplementation` under the mock. We should not need the `beforeEach`.

packages/ui/__tests__/Radio.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { mountAndCheckA11Y } from '@hazelcast/test-helpers'
+
+import { v4 as uuid } from 'uuid'
+import { Radio } from '../src/Radio'
+
+jest.mock('uuid')
+
+const uuidMock = uuid as jest.Mock<ReturnType<typeof uuid>>
+
+describe('Radio', () => {
```
jankoritak, line 11: Can we add these tests?
1. Radio renders label and and input. (Just to make sure everything expected renders).
2. Radio renders helperText.

jankoritak (commented):
Great job overall! Leaving bunch of comments, mostly nits.

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,89 @@
...
+.wrapper {
+  display: inline-flex;
+  // inline-flex changes vertical align compared to display: flex
+  // setting vertical-align allows us to easily use radio button
+  // inline e.g. next to the text
+  // Most common use case will be still the usage in a form where parent
+  // wrapper controls this.
+  vertical-align: bottom;
+  flex-direction: row;
+  user-select: none;
```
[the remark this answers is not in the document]
jvorcak: This was also discussed with me and @aigoncharov before, it's a matter of preference and I've found quite a lot of tutorials where they do it, but since this is the second question about it I suggest we remove it and see how it goes.

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,89 @@
...
+
+.wrapper {
+  display: inline-flex;
+  // inline-flex changes vertical align compared to display: flex
+  // setting vertical-align allows us to easily use radio button
+  // inline e.g. next to the text
+  // Most common use case will be still the usage in a form where parent
+  // wrapper controls this.
+  vertical-align: bottom;
+  flex-direction: row;
```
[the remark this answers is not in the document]
jvorcak: I know it's default behaviour, but I wanted to keep it explicit. I'm fine with removing it thought.

packages/ui/__tests__/Radio.test.tsx
jvorcak: @jankoritak I'll do that thanks!

packages/ui/__tests__/Radio.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+        name="hello"
+        disabled
+        value="world"
+        helperText="This is a helper text."
+        onChange={onChange}
+        onBlur={onBlur}
+        label="Hello World"
+      />,
+    )
+    expect(wrapper.find(Help).exists()).toBeTruthy()
```
danad02, line 56: nit: we can check if we pass `helperText` value to `Help` props.

packages/ui/__tests__/RadioGroup.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+    const wrapper = await mountAndCheckA11Y(
+      <RadioGroup>
+        <Radio checked name="hello" value="hello" disabled onChange={onChange} label="World" />
+        <Radio name="hello" value="world" onChange={onChange2} label="World" />
+      </RadioGroup>,
+    )
+
+    expect(wrapper.find("input[value='hello']").getDOMNode<HTMLInputElement>().disabled).toBe(true)
+    expect(wrapper.find("input[value='world']").getDOMNode<HTMLInputElement>().disabled).toBe(false)
```
danad02, line 21: nit: Should we somehow generate `data-test` props for each `Radio`? 
E.g.:
```
const dataTest = `${value}-radio-data-test`
```

packages/ui/src/Radio.tsx
```diff
@@ -0,0 +1,83 @@
...
+export type RadioExtraProps = {
+  label: string
+  helperText?: string
+  disabled?: boolean
+  required?: boolean
+  className?: string
+  classNameLabel?: string
+}
+
+export type RadioProps = RadioCoreProps & RadioExtraProps & DataTestProp
```
danad02, line 25: We can pass `data-test` prop but we're not using it anywhere.

packages/ui/src/Radio.tsx
```diff
@@ -0,0 +1,83 @@
...
+            [styles.disabled]: disabled,
+          },
+          classNameLabel,
+        )}
+        htmlFor={idRef.current}
+      >
+        <span className={styles.name}>{label}</span>
+        <input
+          type="radio"
+          ref={inputRef}
```
danad02, line 67: Why do we need it?

packages/ui/__tests__/Radio.test.tsx
jvorcak: 🚀 thanks 👏

packages/ui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+        role="radiogroup"
+        className={classNames([styles.radioGroup, radioGroupClassName], {
+          [styles.error]: !!error,
+        })}
+        aria-invalid={!!error}
+        aria-errormessage={error && errorId(idRef.current)}
+      >
+        {children}
+      </div>
+      <Error error={error} className={classNames(styles.errorContainer)} inputId={idRef.current} />
```
danad02, line 34: Should it be `inputId={errorId(idRef.current)}` instead?

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,93 @@
...
+  vertical-align: bottom;
+  cursor: pointer;
+
+  @include c.typographyBodySmall;
+
+  // hide native checkbox
+  input {
+    height: 1px;
+    width: 1px;
+    opacity: 0;
```
danad02, line 20: nit: It's probably not breaking anything but should we also get rid of default `margin`?

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,93 @@
...
+@use '../styles/helpers' as h;
+
+.wrapper {
+  display: inline-flex;
+  // inline-flex changes vertical align compared to display: flex
+  // setting vertical-align allows us to easily use radio button
+  // inline e.g. next to the text
+  // Most common use case will be still the usage in a form where parent
+  // wrapper controls this.
+  vertical-align: bottom;
```
danad02, line 11: Should we also vertically center radio input, label and help icon?
```suggestion
  vertical-align: bottom;
  align-items: center;
```

packages/ui/src/Radio.module.scss
jvorcak: @danad02 we can, but considering other properties it really makes no difference (visual nor semantic). TBH I've never seen it anywhere.

packages/ui/src/Radio.module.scss
jvorcak: Perfect!

packages/ui/src/RadioGroup.tsx
jvorcak: Good catch!

packages/ui/src/Radio.module.scss
danad02: Let's leave it as is then.

packages/ui/__stories__/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+        label="Test"
+        checked={value === 'test'}
+        helperText={'Hello, this is a description'}
+        onChange={({ target: { value } }) => setValue(value)}
+      />
+    </RadioGroup>
+  )
+}
+
+export const RadioGroupWrappedInFormik = () => {
```
aigoncharov, line 65: Great idea!

packages/ui/__stories__/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+        label="Test"
+        checked={value === 'test'}
+        onChange={({ target: { value } }) => setValue(value)}
+      />
+    </RadioGroup>
+  )
+}
+
+export const RadioGroupWithDescriptionAndError = () => {
+  const idRef = useRef(uuid())
```
aigoncharov, line 40: nit: Couldn't we use a simple string top-level constant in the beginning of the file instead?

packages/ui/src/Checkbox.module.scss
```diff
@@ -23,12 +22,22 @@
...
+    width: 1px;
     opacity: 0;
     position: absolute;
```
aigoncharov, line 28: nit
```suggestion
    @include h.visuallyHidden;
```

packages/ui/src/Checkbox.tsx
```diff
@@ -26,7 +25,7 @@ export type CheckboxExtraProps = {
   classNameLabel?: string
 }
 
-type CheckboxProps = CheckboxCoreProps & CheckboxExtraProps & DataTestProp
```
aigoncharov, line 29: I would suggest leaving `DataTestProp`, and pass its value to the top-level `div`. We use it in MC to find the elements in E2E tests.

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,94 @@
...
```
aigoncharov, line 11: If we are considering the inline usage next to the text, should we add a story to capture it?

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,94 @@
...
+  cursor: pointer;
+
+  @include c.typographyBodySmall;
+
+  // hide native checkbox
+  input {
+    height: 1px;
+    width: 1px;
+    opacity: 0;
+    position: absolute;
```
aigoncharov, line 22: nit
```suggestion
    @include h.visuallyHidden;
```

packages/ui/src/Radio.tsx
```diff
@@ -0,0 +1,83 @@
...
+  label,
+  required,
+  helperText,
+  disabled = false,
+  checked,
+}) => {
+  const idRef = useRef(uuid())
+
+  return (
+    <span className={className}>
```
aigoncharov, line 47: Why do we need this top-level span?

packages/ui/src/Radio.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { ChangeEvent, FC, FocusEvent, useRef } from 'react'
+import styles from './Radio.module.scss'
+import classNames from 'classnames'
+import { Help, helpTooltipId } from './Help'
+import { v4 as uuid } from 'uuid'
+
+type RadioCoreProps = {
+  name: string
```
aigoncharov, line 8: `Radio` does not exist outside of `RadioGroup`, right? Moreover, all `Radio` elements inside of the same group have the same name. Couldn't we create a context inside of the `RadioGroup`, and make the `Radio` consume the `name` from the context? It would save us the error of when we supply different names to the radios of the same group by mistake. WDYT?

packages/ui/src/RadioGroup.module.scss
```diff
@@ -0,0 +1,12 @@
+@use '../styles/constants' as c;
+@use '../styles/helpers' as h;
+
+.radioGroup {
+  display: flex;
+  flex-direction: column;
+
+  // Error state
+  &.error label > input:checked ~ * {
```
aigoncharov, line 9: We are kind of violating the incapsulation of the `Radio` component here. Building on the context idea (https://github.com/hazelcast/frontend-shared/pull/54/files#r517947676), couldn't we pass the `error: boolean` to the context and make the `Radio` style the error state internally based on the prop?

packages/ui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * ### Purpose
+ * Forms require input from users. If you need information that can be represented as a choice of multiple values, use input with type 'radio'.
+ * You can use RadioGroup in order to group radios buttons into logical cells and provide corresponding error if needed.
+ */
+export const RadioGroup: FC<RadioGroupProps> = ({ radioGroupClassName, error, children }) => {
+  const idRef = useRef(uuid())
+
+  return (
+    <div>
```
aigoncharov, line 23: nit: Should we also allow to style this parent div?

packages/ui/src/Radio.tsx
jvorcak: I tried to support it outside, but I'm just trying to make things as similar to native HTML as possible. I should probably stop doing it and start doing it more pragmatic. Context sound great!

packages/ui/__stories__/RadioGroup.stories.tsx
jvorcak: @aigoncharov since this ID is used as a name to a native HTML radio input, we don't want different stories to re-use it because it'd link different stories together.

packages/ui/src/Checkbox.tsx
jvorcak: @aigoncharov, good. Should we then do this for all of the components?

aigoncharov: I would say so, yeah.

packages/ui/src/Radio.tsx
jvorcak: Removed

packages/ui/src/RadioGroup.tsx
danad02: @jvorcak I think you changed it back unintentionally.
```suggestion
      <Error error={error} className={classNames(styles.errorContainer)} inputId={errorId(idRef.current)} />
```

jvorcak: @danad02 not really, If you have a look at `Error` implementation, it receives `inputId` as an argument and does ` id={errorId(inputId)}` so in your case I'd add `-error` string twice.

danad02: You're right, sorry about that!

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,105 @@
...
+    border-radius: 50%;
+    // because of ~ and + selectors used below, we need to keep the current order in DOM and
+    // fix it visually with order property
+    order: -1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    // in case we have inline mode, we should have a padding like native radio button has
+    margin-left: c.$grid;
```
danad02, line 38: This also adds margin to default (stacked) radios:
image

jvorcak: @danad02 should we just forget about it and leave this to the user who will use inline styling? This will be the most common you case.

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  .name {
+    padding-left: c.$grid * 1.25;
+    padding-right: c.$grid * 1.25;
+  }
+
+  // Error state
+  &.error input:checked ~ * {
+    border-color: c.$colorError;
```
danad02, line 76: This also changes border color for Help tooltip:
image

jvorcak: Nice catch! Fixed

packages/ui/__stories__/RadioGroup.stories.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useRef, useState } from 'react'
+import { v4 as uuid } from 'uuid'
+import { Radio } from '../src/Radio'
+import { logger } from '@hazelcast/services'
+import { RadioGroup } from '../src/RadioGroup'
+import { Form, Formik } from 'formik'
+import { RadioGroupFieldFormik } from '../src/RadioGroupFormik'
+import { RadioFieldFormik } from '../src/RadioFormik'
+
```
danad02, line 9: Can we also have an `inline` radio group story?

jvorcak: @danad02 we do have it, just in a radio story. They're quite related and the only reason they're separated is so that we have a documentation for both of them available. I can move the inline scenario to RadioGroup stories though.

danad02: @jvorcak Moving it here would make sense, thanks :+1:

packages/ui/src/Radio.module.scss
danad02: @jvorcak I have to admit that I'm not a fan of the solution you propose. I suggest following:
Add boolean type `inline` prop to `RadioGroup` component. If it's true then we use `flex-direction: row`, otherwise `flex-direction: columns`. WDYT?

jvorcak: I thought about it, but I wanted to avoid us having to pass a property that has only to do sth with styling and not the actual functionality, but I guess this is a corner case where we can do that and it'll be easier to use this OOTB. It's fine with me.

danad02: @jvorcak You don't have to anything with context.

danad02: Everything you would need to do is:
```
export type RadioGroupProps = RadioGroupCoreProps & {
  ...
  inline?: boolean
} & DataTestProp

/**
 * ### Purpose
 * Forms require input from users. If you need information that can be represented as a choice of multiple values, use input with type 'radio'.
 * You can use RadioGroup in order to group radios buttons into logical cells and provide corresponding error if needed.
 */
export const RadioGroup: FC<RadioGroupProps> = ({
  ...
  inline = false,
}) => {
  const idRef = useRef(uuid())

  return (
    <div className={className} data-test={dataTest}>
      <div
        role="radiogroup"
        data-test="radio-group"
        className={classNames([
          styles.radioGroup,
          radioGroupClassName,
          {
            [styles.inline]: inline,
          },
        ])}
        aria-invalid={!!error}
        aria-errormessage={error && errorId(idRef.current)}
      >
        <RadioGroupContext.Provider value={{ name, error, onChange }}>{children}</RadioGroupContext.Provider>
      </div>
      <Error error={error} className={classNames(styles.errorContainer)} inputId={idRef.current} />
    </div>
  )
}
```

And in `RadioGroup.module.scss` you would add: 
```
.inline {
  flex-direction: row;
}
```

danad02: But you have to add side margins in children so in the end it would truly require using the context.

jvorcak: @danad02 I've done the version with context, pushing soon after screenshots are updated.

danad02 (approved):
Great job!

packages/ui/src/Radio.tsx
```diff
@@ -0,0 +1,82 @@
...
+  required,
+  helperText,
+  disabled = false,
+  checked,
+  'data-test': dataTest,
+}) => {
+  const idRef = useRef(uuid())
+
+  return (
+    <RadioGroupContext.Consumer>
```
aigoncharov, line 46: nit: Since we are using hooks everywhere I would suggest utilizing `useContext` instead

packages/ui/src/Radio.tsx
```diff
@@ -0,0 +1,82 @@
...
+      {({ name, onChange, error, inline }) => (
+        <label
+          className={classNames(styles.wrapper, className, {
```
aigoncharov, line 49: nit: I would suggest putting `className` as the last item, so it could override anything

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,109 @@
...
+    align-items: center;
+    justify-content: center;
+  }
+
+  &.inline {
+    .checkmark {
+      // in case we have inline mode, we should have a padding like native radio button has
+      margin-left: c.$grid;
+    }
+  }
```
aigoncharov, line 43: What if there is no leading text before the group of the radios? Wouldn't the margin feel weird then? I would suggest adding something like:
```
.radioGroup {
  display: flex;
  flex-direction: column;

  &.inline > *:not(:first-child) {
    margin-left: c.$grid;
  }
}
```
to the radio group instead. Then we would be able to remove the `inline` from the context.
I see why you wanted to add the `inline` to the context just like we did with the `error`. However, in my mind they are two different beasts. When we styled the error from the radio group we relied on the **implementation detail** of the radio. In other words, we had to know how it is implemented. And we affected that implementation from the outside.
With the margin, we are styling children's spacing as black boxes. We do not change their internal implementation. 
Does it make my sense? WDYT?

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+    Disabled
+  */
+  &.disabled {
+    cursor: not-allowed;
+    color: c.$colorNeutralLight;
+  }
+
+  &.focus > input:not(:disabled),
```
aigoncharov, line 91: nit: AFAIK, browsers do not allow focusing disabled elements, so we could safely remove `:not(:disabled)` from here.

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,109 @@
...
+  // inline e.g. next to the text
+  // Most common use case will be still the usage in a form where parent
+  // wrapper controls this.
+  vertical-align: bottom;
+  align-items: center;
+  cursor: pointer;
+
+  @include c.typographyBodySmall;
+  // we need to provide spacing between radios if they're in a vertical column
+  margin-bottom: c.$grid;
```
aigoncharov, line 17: Isn't this spacing going to add the bottom margin to the last radio? Do we always want to do it? Should we consider something in a fashion of https://github.com/hazelcast/frontend-shared/pull/54/files#r519910049?

packages/ui/src/Radio.module.scss
```diff
@@ -0,0 +1,109 @@
...
+    border-color: c.$colorError;
+  }
+
+  /**
+    Disabled
+  */
+  &.disabled {
+    cursor: not-allowed;
+    color: c.$colorNeutralLight;
+  }
```
aigoncharov, line 89: nit
```suggestion
  &.disabled {
    cursor: not-allowed;
    color: c.$colorNeutralLight;
    
    input ~ .checkmark {
      border-color: c.$colorNeutralLight;
    }
    input:checked ~ .checkmark {
      border-color: c.$colorNeutral;
    }
  }
```
Then we could remove all from L98. Feels a bit more concise

packages/ui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+        role="radiogroup"
+        data-test="radio-group"
+        className={classNames(
+          {
+            [styles.inline]: inline,
+          },
+          [styles.radioGroup, radioGroupClassName],
+        )}
+        aria-invalid={!!error}
+        aria-errormessage={error && errorId(idRef.current)}
```
aigoncharov, line 51: On Ubuntu the default screen reader does not read the error message out loud if it is linked to the radio group. Could you check it on Mac?

I would suggest passing `errorId?: string` to the context. Then we would be able to apply `aria-invlid` and `aria-errormessage` to every radio. Maybe, it would help with the screen reader. WDYT?

packages/ui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+      >
+        <RadioGroupContext.Provider value={{ name, error, onChange, inline }}>{children}</RadioGroupContext.Provider>
```
aigoncharov, line 53: nit: Should we memoize the provider value?

packages/ui/src/RadioFormik.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { ReactElement, useContext } from 'react'
+import { FieldValidator, useField } from 'formik'
+
+import { Radio, RadioExtraProps } from './Radio'
+import { RadioGroupContext } from './RadioGroupContext'
+
+export type RadioFormikProps = RadioExtraProps & {
+  validate?: FieldValidator
```
aigoncharov, line 8: Why do we need a field validator here? We already have one at the `RadioGroupFormik` level. Even if this field validator errors the error would not be handled and displayed by us, would it?

packages/ui/src/Radio.tsx
jvorcak: This is actually not a `nit` I think.

packages/ui/src/Radio.module.scss
jvorcak: I actually prefer to have it there for consistency, I like to think of it as part of the element and when I placed some text under a radio group it felt natural to me. Don't you think we could leave it as is?

packages/ui/src/Radio.module.scss
jvorcak: I never liked passing it through the context, but just to make sure we realise this, this kills the ability of having intermediate DOM elements in between RadioGroup and Radio elements. But I'm fine with that.

packages/ui/src/Radio.module.scss
jvorcak: There are corner cases where it wouldn't work (e.g. disabled and checked). But I've added it and made it more SCSS-like.

packages/ui/src/RadioGroup.tsx
jvorcak: It reads it for me, what I could do is to pass errorId instead of error message right now and apply to only for checked button?

packages/ui/src/RadioGroup.tsx
jvorcak: Is it really necessary? Memoize based on what?

danad02: @jvorcak I believe @aigoncharov means to do something like this:
```
const providerValue = useMemo(() => ({ name, error, onChange, inline }), [name, error, onChange, inline])

<RadioGroupContext.Provider value={value}>{children}</RadioGroupContext.Provider>
```

packages/ui/src/Radio.module.scss
danad02: @jvorcak How exactly does this kill our Radio Inline scenario? What kind of other elements would you like to insert there?

packages/ui/src/Radio.module.scss
aigoncharov: @jvorcak I get the sane defaults argument. I am just a little uncomfortable with the lest radio always having the bottom margin.
@danad02 @jankoritak @jozefhruska WDYT?
Personally, I feel like adding top-level margins affects the external layout - the layout of the parent container and it, in my view, kind of, breaks component encapsulation. My mental model is that each component should take care of its visuals and its internal layout only, leaving the external layout to the parent. However, I am open to being convinced otherwise. After all, we all will be using this lib. We all should be comfortable with it.

packages/ui/src/RadioGroup.tsx
aigoncharov: @jvorcak ^^^

packages/ui/src/Radio.module.scss
aigoncharov: @jvorcak I feel like we do not have too many options here:
1. Use the `inline` prop in the context. Pro: Radio will be able to add margin on its own. Con: Radio will always have the margin, even it is the first Radio in the row. Additionally, it kind of breaks encapsulation see https://github.com/hazelcast/frontend-shared/pull/54#discussion_r520589718
2. Let the parent handle the margin. Agree that it will work only if Radios are direct descendants. However, the user can always override the class for more complex scenarios. And are there any complex scenarios in the first place?

jvorcak: Yeah, I agree with number 2 and that is the current implementation, so you can resolve if you're happy with it.

packages/ui/src/Radio.module.scss
jvorcak: This is how it looks with and without it.
Screenshot 2020-11-10 at 16 12 44
Screenshot 2020-11-10 at 16 12 48

danad02: I think it would be better for parent (`RadioGroup` in this case) to handle this margin.

jvorcak: https://github.com/hazelcast/frontend-shared/pull/54/commits/5ac6232bfc914d761f7501e0551ed77dde24a7bc

packages/ui/src/RadioGroup.module.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  &.inline {
+    flex-direction: row;
+
+    > *:not(:first-child) {
+      margin-left: c.$grid;
+    }
+  }
+}
```
aigoncharov, line 20:
```suggestion
}

.errorContainer {
  margin-top: c.$grid;
}
```

packages/ui/src/Radio.module.scss
aigoncharov: @jvorcak this small proposal could help us restore the spacing for the error https://github.com/hazelcast/frontend-shared/pull/54/files#r521216637

packages/ui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,63 @@
...
+          },
+          styles.radioGroup,
+          radioGroupClassName,
+        )}
+        aria-invalid={!!error}
+        aria-errormessage={errorIdString}
+      >
+        <RadioGroupContext.Provider value={providerValue}>{children}</RadioGroupContext.Provider>
+      </div>
+      <Error error={error} className={classNames(styles.errorContainer)} inputId={idRef.current} />
```
aigoncharov, line 60:
```suggestion
      <Error error={error} className={styles.errorContainer} inputId={idRef.current} />
```

aigoncharov (approved):
Great job! Thank you for your patience!

packages/ui/src/Checkbox.module.scss
```diff
@@ -2,13 +2,17 @@
...
-  user-select: none;
   cursor: pointer;
 
   @include c.typographyBodySmall;
 
+  // hide native checkbox
+  input {
+    @include h.visuallyHidden;
+  }
+
```
aigoncharov, line 15: Don't we need it since we changed the display value from `inline-flex` to `flex`?
```suggestion

.errorContainer {
  margin-top: c.$grid;
}
```
I think it is the last comment. Feel free to merge the PR in once it is resolved.

PR #60, opened 2020-11-03
Fix the disabled button tooltip

### LINKED ISSUE #58: Disabled button does not render a tooltip
Try
1. Make button relative
2. Create an absolutely positioned child:
   ```
    position: absolute;
    opacity: 0;
    pointer-events: none;
    left:0;
    top: 0;
    width: 100%;
    height: 100%;
    ```
3. Link the tooltip to the absolutely positioned child

aigoncharov:
Closes #58


jankoritak (approved):
Thanks for taking care of this!

PR #61, opened 2020-11-04
feat: badge component

### LINKED ISSUE #12: [UI core library] Badge
https://zeroheight.com/11d0e6dac/p/02869f-badge

1. Create a `TextBadge` component with API
    1. content (as children)
    2. type
2. Create a `NumberBadge` component with API
    1. value

jankoritak:
Closes #12 

#### Notes:
- [x] Colours need to be iterated with @pawelklasa, as the current scale do not provide enough contrast


packages/ui/src/Badge.tsx
```diff
@@ -0,0 +1,71 @@
...
+    icon: AlertCircle,
+    ariaLabel: 'Info critical circle icon',
+  },
+}
+
+export type BadgeSize = 'small' | 'normal'
+
+export type BadgeProps = {
+  type: BadgeType
+  size?: BadgeSize
```
jvorcak, line 43: GJ @jankoritak 👏 

I have one concern about this specific property. Do we want it? Shouldn't we leave it to the user to decide what is the font-size and inherit it? + define vertical-align to baseline?

I like how Bootstrap does it and it feels more natural https://getbootstrap.com/docs/4.5/components/badge/

We use these badges exactly like this in .cloud (part of some headings etc. and I'd like to avoid changing the property every time I change the semantics of h1-h6. I think it should work OOTB in the headings - regular text. 
WDYT?

jvorcak (changes requested):

packages/ui/__stories__/NumberBadge.stories.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { NumberBadge } from '../src/NumberBadge'
+
+export default {
+  title: 'components/NumberBadge',
+  component: NumberBadge,
+}
+
+export const Default = () => <NumberBadge number={1} />
+
+export const LessThan99 = () => <NumberBadge number={99} />
```
danad02, line 12: nit: 99 is not lesser than 99 😄 
```suggestion
export const LessThan100 = () => <NumberBadge number={99} />
```

packages/ui/__tests__/NumberBadge.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+  })
+
+  it('Renders formatted number higher than 99', async () => {
+    const number = 100
+
+    const wrapper = await mountAndCheckA11Y(<NumberBadge number={number} />)
+
+    expect(wrapper.findDataTest('number-badge').text()).toBe('+99')
+  })
+})
```
danad02, line 21: nit: I would go with number of digits in descriptions:
```suggestion
describe('NumberBadge', () => {
  it('Renders two digit number', async () => {
    const number = 99

    const wrapper = await mountAndCheckA11Y(<NumberBadge number={number} />)

    expect(wrapper.findDataTest('number-badge').text()).toBe(number.toString())
  })

  it('Renders three digit number', async () => {
    const number = 100

    const wrapper = await mountAndCheckA11Y(<NumberBadge number={number} />)

    expect(wrapper.findDataTest('number-badge').text()).toBe('+99')
  })
})
```

packages/ui/__stories__/NumberBadge.stories.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export const MoreThan99 = () => <NumberBadge number={100} />
```
danad02, line 14: nit: How about this?
```suggestion
export const OneDigit = () => <NumberBadge number={1} />

export const TwoDigits= () => <NumberBadge number={99} />

export const ThreeDigits = () => <NumberBadge number={100} />
```

packages/ui/__tests__/NumberBadge.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { mountAndCheckA11Y } from '@hazelcast/test-helpers'
+import React from 'react'
+import { NumberBadge } from '../src/NumberBadge'
+
+describe('NumberBadge', () => {
+  it('Renders number lower than 99', async () => {
```
danad02, line 6:
```suggestion
  it('Renders number lower than 100', async () => {
```

packages/ui/__stories__/NumberBadge.stories.tsx
danad02, line 12:
```suggestion
export const LessThan100 = () => <NumberBadge number={99} />
```

packages/ui/src/Badge.tsx
danad02: That's a good idea 👍 It would certainly feel more natural.

jankoritak: We can do that.

However, this component currently always renders an icon with the content, does that work for your use-case?

jvorcak: We'll see about how it first with the future design, however, an icon is just SVG so i could follow the same rules?

jankoritak: Sure, we can make the inheritance work, I'm just curious about the icon, as Bootstrap's Badge looks a bit different.

jankoritak: Addressed in: https://github.com/hazelcast/frontend-shared/pull/61/commits/59ef018563f9192c571e9c62eafde80439a8f206. Please, re-review. @jvorcak

aigoncharov: Guys, I am with @jankoritak on this one. Our icons come with two different sizes. I do not think we should scale them dynamically like the font size. 
On the other hand, I see the point @jvorcak is trying to make.
@jvorcak do you have places where you want a non-standard font-size for the badge? In MC they are always of the same size no matter where we place them.

aigoncharov: There is an option to resolve the icon size implicitly. We could inherit the font-size for the text. Then we could:
1. Run `useLayoutEffect` on the first render only.
2. Calculate the inherited `font-size` there via https://stackoverflow.com/questions/1955048/get-computed-font-size-for-dom-element-in-js
3. Based on the calculated `font-size` pass one of our existing icon sizes to the `Icon` component.
But I think we should have a valid use-case to complicate the component with this logic.

jvorcak: @jankoritak @aigoncharov It'd mark this as nice-to-have, I can live with 'n' static sizes if that suites your use-case better.

The reason why I wanted this is that badges (with and without icons) are often used in headings (h1-h6) and I expect to look nice OOTB. Since an icon is SVG and the rest of the component is just text, I see no reason why we would need JS for this.

I try to see how other component libs solve the issues we have, Bootstrap handles this as I showed earlier, which I think is quite elegant. *On the other hand* Material-ui has a [chip](https://material-ui.com/components/chips/#chip) component with defined static sizes. So it really depends on how do we want to use it.

jvorcak: It seems other libs sometimes even have two options, depending on whether this is semantically used as a standalone component or just an addition to a text. We don't have a lot of cases like this on .cloud, 1-3, so please feel free to pick whichever suits your need in a better way.

aigoncharov: @jankoritak @jvorcak I would suggest sticking to static sizes until the need of the dynamic inherited size arises. Once it does, we will add the `inherit` size prop for that use-case. WDYT?

jankoritak: As for scaling of the icon, we would not need JS there, the current state (using `em`) works just fine. Although it may look a bit confusing, as we're overriding the `size` prop of the `Icon` via CSS.

I see we have a consensus now though, let me roll back to static sizes then.

danad02 (approved):
LGTM!

jvorcak (approved):

packages/ui/src/NumberBadge.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { FC } from 'react'
+
+import styles from './NumberBadge.module.scss'
+
+export type NumberBadgeProps = {
+  number: number
```
aigoncharov, line 6: nit: Should we also add `size` here?

packages/ui/__stories__/Badge.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import { Badge } from '../src/Badge'
+
+export default {
+  title: 'components/Badge',
+  component: Badge,
+}
+
+export const NeutralNormal = () => <Badge size="normal" type="neutral" content="Badge text" />
```
aigoncharov, line 10: nit: It seems like we support multi-line content. Should we add at least one story with the multi-line content?

aigoncharov (approved):
Great job!

packages/ui/src/NumberBadge.tsx
jankoritak: There is only one site in ZeroHeight. AFAIK, we'll only use the `NumberBadge` for side menu, so there should not be need for that at the moment.

PR #62, opened 2020-11-05
Feature/textarea

### LINKED ISSUE #8: [UI core library] Text Area
https://zeroheight.com/11d0e6dac/p/316944-text-field/t/724564

        1. name
        2. value
        3. onChange
        4. onBlur
        5. label
        6. error
        7. touched
2. Create a formik v2 wrapper

jankoritak:
Closes #8

#### Description
- As we don't have explicit Figma design for text area, it's mostly a clone of text field.


packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,80 @@
...
+  className,
+  error,
+  inputClassName,
+  label,
+  name,
+  onBlur,
+  onChange,
+  placeholder,
+  value,
+  defaultValue,
```
aigoncharov, line 38: Why do we need the `defaultValue`?

aigoncharov, line 38: Do you think we will need to use the textarea as an uncontrolled component?

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,80 @@
...
+export type TextAreaExtraProps = {
+  label: string
+  inputClassName?: string
+  errorClassName?: string
+  inputContainerChild?: ReactElement
+} & Partial<Pick<HTMLTextAreaElement, 'className' | 'defaultValue' | 'disabled' | 'placeholder' | 'required'>>
+
+export type TextAreaProps = TextAreaCoreProps & TextAreaExtraProps & DataTestProp
+
+export const TextArea: FC<TextAreaProps> = ({
```
aigoncharov, line 28: I would also suggest adding `resizable?: boolean` prop

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onBlur?: (e: FocusEvent<HTMLTextAreaElement>) => void
+  onChange: (e: ChangeEvent<HTMLTextAreaElement>) => void
+  error?: string
+}
+
+export type TextAreaExtraProps = {
+  label: string
+  inputClassName?: string
+  errorClassName?: string
+  inputContainerChild?: ReactElement
```
aigoncharov, line 23:
```suggestion
  errorClassName?: string
```

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { ChangeEvent, FC, FocusEvent, ReactElement, useRef } from 'react'
+import { v4 as uuid } from 'uuid'
+import cn from 'classnames'
+
+import { DataTestProp } from '@hazelcast/helpers'
+import { HiddenLabel } from '../src/HiddenLabel'
+import { Error, errorId } from '../src'
```
aigoncharov, line 7:
```suggestion
import { Error, errorId } from '../src/Error'
```

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,80 @@
...
+  name: string
+  value?: string
+  onBlur?: (e: FocusEvent<HTMLTextAreaElement>) => void
+  onChange: (e: ChangeEvent<HTMLTextAreaElement>) => void
+  error?: string
+}
+
+export type TextAreaExtraProps = {
+  label: string
+  inputClassName?: string
```
aigoncharov, line 21: nit
```suggestion
  textareaClassName?: string
```

packages/ui/src/TextArea.tsx
jankoritak: Good point, didn't realise that, thanks!

packages/ui/src/TextArea.tsx
jankoritak: Good idea!

jankoritak:
Ready for re-review.

danad02:
@jankoritak Could you merge master and add label?

danad02:
@jankoritak Can you merge master and update label?

packages/ui/src/TextArea.module.scss
```diff
@@ -0,0 +1,89 @@
...
+  box-sizing: border-box;
+  margin: 0;
+
+  .label {
+    margin-bottom: c.$grid;
+  }
+
+  .helperText {
+    display: block;
+    margin: c.$grid * 2 0 0 c.$grid * 2;
```
aigoncharov, line 17: Help also has internal padding of `c.$grid`. So with this margin we have `3*c.$grid` left spacing in total. For the TextField  it is only `2*c.$grid`. Do you think it would make sense to match the TextField spacing here as well?
```suggestion
    margin: c.$grid 0 0 c.$grid;
```

packages/ui/src/TextArea.module.scss
```diff
@@ -0,0 +1,89 @@
...
+    letter-spacing: inherit;
+    color: inherit;
+    border-radius: c.$borderRadius;
+    background-color: c.$colorNeutralLighter;
+
+    &::placeholder {
+      color: c.$colorTextSecondary;
+    }
+
+    // Note: Resize is not supported by IE11, it's presence does no harm though
```
aigoncharov, line 48: A great example of graceful degradation!

packages/ui/src/TextArea.module.scss
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  &.hasError {
+    textarea {
+      border-color: c.$colorError;
+    }
+  }
+
+  textarea:focus {
```
aigoncharov, line 81: I am wondering if we have to make the outline a bit more complicated because of https://hazelcast.slack.com/archives/C01BT11F24E/p1605077912095500
I would suggest to try:
1. Add an absolute positioned element to the .textAreaContainer
2. Use `:focus-within` on that element to draw the outline there
3. Leave the `:focus` styling of the outline for IE11 as it does not support `:focus-within`
4. To make it disappear in other browser add `:focus` override for the parent that have `:focus-within`

WDYT?

packages/ui/src/TextArea.module.scss
```diff
@@ -0,0 +1,89 @@
+@use '../styles/constants' as c;
```
aigoncharov, line 1: Great job on the styles! Very readable!c Love the nesting!

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onBlur?: (e: FocusEvent<HTMLTextAreaElement>) => void
+  onChange: (e: ChangeEvent<HTMLTextAreaElement>) => void
+  error?: string
+}
+
+export type TextAreaExtraProps = {
+  label: string
+  textareaClassName?: string
+  errorClassName?: string
+  resizable?: boolean
```
aigoncharov, line 24: nit: Should we have a story for the non-resizable text area?

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,89 @@
...
+      <Label id={idRef.current} label={label} />
+      <div className={styles.textAreaContainer}>
+        <textarea
+          aria-invalid={!!error}
+          aria-required={required}
+          aria-errormessage={error && errorId(idRef.current)}
+          id={idRef.current}
+          className={cn(textareaClassName, {
+            [styles.notResizable]: !resizable,
+          })}
```
aigoncharov, line 72: Should we place the overriding className as the last item in the list so it could override anything?
```suggestion
          className={cn({
            [styles.notResizable]: !resizable,
          }, textareaClassName)}
```

packages/ui/__tests__/TextArea.test.tsx
```diff
@@ -0,0 +1,275 @@
+import React from 'react'
```
aigoncharov, line 1: Good test coverage!

packages/ui/src/TextAreaFormik.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ReactElement } from 'react'
```
aigoncharov, line 1: Should we have a test for the formik-integrated component?

packages/ui/src/TextArea.module.scss
jankoritak: Good catch!

packages/ui/src/TextArea.tsx
jankoritak: We already do. 👍

packages/ui/__stories__/TextArea.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+    value={text}
+    placeholder="Describe character"
+    label="Character description"
+    onBlur={() => logger.log('blur')}
+    onChange={(e) => logger.log('change', e.target.value)}
+    resizable={false}
+  />
+)
+
+export const WithHelperText = () => (
```
jvorcak, line 111: When focusing a textarea with helper text I see two focus borders. + if I hover over help, I don't expect hover to be activated.

Screenshot 2020-11-12 at 10 08 59
Screenshot 2020-11-12 at 10 08 56

jvorcak (dismissed):

packages/ui/__stories__/TextArea.stories.tsx
aigoncharov: Quite a catch!

danad02: The two outlines appear every time if you try to make text area smaller in width.
image

danad02: The scrollbar is overlapping round corner borders on chrome, linux:
image
image

aigoncharov: Fixed in https://github.com/hazelcast/frontend-shared/pull/62/commits/356c037e0d2083ee516ebbf607e8bf56de0eb669

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+            name={name}
+            onBlur={onBlur}
+            onChange={onChange}
+            required={required}
+            value={value}
+            disabled={disabled}
+            {...htmlAttrs}
+          />
+        </div>
+        {helperText && (
```
jvorcak, line 88: We have a problem when resizing an element. The popup appears at the original place all the time. It's very possible that it's not a problem of this component.

Screenshot 2020-11-12 at 10 34 47

packages/ui/src/TextArea.module.scss
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    &::placeholder {
+      color: c.$colorTextSecondary;
+    }
+  }
+
+  &.empty {
+    color: c.$colorTextSecondary;
+  }
```
danad02, line 82: I didn't notice it in `TextField` but why do we need it? Isn't `placeholder` selector enough?

danad02 (approved):
Great job! @aigoncharov Please, merge after dealing with gaps between border and inset box shadow.

packages/ui/src/TextArea.tsx
aigoncharov: I have just found it as well. Working on the solution.

packages/ui/src/TextArea.module.scss
aigoncharov: For the TextField we need to color the icon inside of it. Here we do not need it indeed. Nice catch!

packages/ui/src/TextArea.tsx
```diff
@@ -0,0 +1,95 @@
...
+  error?: string
+}
+
+export type TextAreaExtraProps = {
+  label: string
+  textareaClassName?: string
+  errorClassName?: string
+  resizable?: boolean
+  helperText?: string | ReactElement
+} & Partial<Pick<HTMLTextAreaElement, 'className' | 'disabled' | 'placeholder' | 'required'>>
```
jvorcak, line 26: Should we also support `rows`? So that we can have control of an initial height?

aigoncharov: Do you use rows anywhere in the Cloud? I suggest we make our TextArea of fixed height which we can override with a className. WDYT?

jvorcak: I think it's nice HTML standard attr that should be allowed, but we're not heavily dependent on it. It's more like `nit`

danad02 (approved):
Great job! It was tough but you've done it guys. Cheers :tada:

jankoritak:
✅ I can't provide a review, but LGTM! Thanks for finishing the PR!

PR #63, opened 2020-11-06
NumberField

### LINKED ISSUE #6: [UI core library] Number field
https://zeroheight.com/11d0e6dac/p/514540-input-stepper

        1. name
        2. value
        3. onChange
        4. onBlur
        5. label
        6. error
        7. touched
2. Create a formik v2 wrapper

aigoncharov:
Closes #6. Makes labels visible.


packages/ui/src/NumberField.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const onIncrement = useCallback(() => {
+    let newValue = value + step
+    if (max !== undefined && newValue > max) {
+      newValue = max
+    }
+    onChange(newValue)
+  }, [value, onChange, step, max])
+
+  const overlay = (
```
danad02, line 75: nit: we can memoize this.

packages/ui/__tests__/NumberField.test.tsx
```diff
@@ -0,0 +1,245 @@
...
+    const onChange = jest.fn()
+
+    const wrapper = await mountAndCheckA11Y(
+      <NumberField name="name" value={42} max={42} placeholder="Enter the name" label="Wisest jedi" onBlur={onBlur} onChange={onChange} />,
+    )
+
+    expect(wrapper.find(IconButton).at(1).prop('disabled')).toBe(true)
+  })
+
+  it('the initial value is adjusted if it is less than min', async () => {
```
danad02, line 232: nit: Can we have this kind of test for `max`?

packages/ui/src/NumberField.tsx
aigoncharov: onIncrement and onDecrement are updated with every value change. So `overlay` is updated as well. I just feel it would be updated too frequently for `useMemo` to truly shine here. WDYT?

danad02: Wouldn't the same logic apply to `useCallback` used for `onDecrement` and `onIncrement`? 
Anyway, you're right about frequent updates. Let's leave it as is.

aigoncharov: You are right. It should apply. Updated!

packages/ui/src/NumberField.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return (
+    <TextField
+      {...props}
+      name={name}
+      value={value}
+      onChange={onChangeWrapped}
+      type="number"
+      inputContainerChild={overlay}
+      inputClassName={cn(styles.inputContainer, className)}
```
danad02, line 118: I believe it should be like this:
```suggestion
  className,
  inputClassName,
  onChange,
  name,
  ...props
}) => {
  useLayoutEffect(() => {
    if (min !== undefined && value < min) {
      onChange(min)
    }
    if (max !== undefined && value > max) {
      onChange(max)
    }
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [min, max])

  const onDecrement = useCallback(() => {
    let newValue = value - step
    if (min !== undefined && newValue < min) {
      newValue = min
    }
    onChange(newValue)
  }, [value, onChange, step, min])

  const onIncrement = useCallback(() => {
    let newValue = value + step
    if (max !== undefined && newValue > max) {
      newValue = max
    }
    onChange(newValue)
  }, [value, onChange, step, max])

  const overlay = useMemo(
    () => (
      <>
        <IconButton
          size="small"
          icon={Minus}
          iconAriaLabel={decrementIconAriaLabel}
          className={styles.decrement}
          onClick={onDecrement}
          disabled={min !== undefined && value <= min}
          kind="primary"
          data-test="number-field-decrement"
        />
        <IconButton
          size="small"
          icon={Plus}
          iconAriaLabel={incrementIconAriaLabel}
          className={styles.increment}
          onClick={onIncrement}
          disabled={max !== undefined && value >= max}
          kind="primary"
          data-test="number-field-increment"
        />
      </>
    ),
    [onIncrement, onDecrement, decrementIconAriaLabel, incrementIconAriaLabel, min, max, value],
  )

  const onChangeWrapped = useCallback(
    (e: ChangeEvent<HTMLInputElement>) => {
      onChange(numberType === 'int' ? parseInt(e.target.value, 10) : parseFloat(e.target.value))
    },
    [onChange, numberType],
  )

  return (
    <TextField
      {...props}
      name={name}
      value={value}
      onChange={onChangeWrapped}
      type="number"
      inputContainerChild={overlay}
      className={className}
      inputClassName={cn(styles.inputContainer, inputClassName)}
```

packages/ui/src/NumberField.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+export const NumberField: FC<NumberFieldProps> = ({
+  incrementIconAriaLabel = 'Increment',
+  decrementIconAriaLabel = 'Decrement',
+  step = 1,
+  min,
+  max,
+  value,
+  numberType = 'int',
+  className,
```
danad02, line 44:
```suggestion
  inputClassName,
  className,
```

packages/ui/src/NumberField.tsx
danad02, line 118: And use the `inputClassName` here:

```suggestion
      inputClassName={cn(styles.inputContainer, inputClassName)}
```

danad02 (approved):
LGTM!

PR #66, opened 2020-11-11
Storybook works with IE11

### LINKED ISSUE #28: Make storybook work in IE11
Maybe, try this https://github.com/storybookjs/storybook/issues/9242#issuecomment-693884172

jvorcak:
Fixes #28 

Maybe we should also include `helpers` folder? Is it indented to be used directly in the components? 
Or we can do that easily once we need it.


aigoncharov (approved):
LGTM! Works with my VM-based IE11

PR #67, opened 2020-11-11
feat(ui/icon): update aria properties

jozefhruska:
The change for Icon props as we discussed, `aria-hidden` is required for use cases similar to this:
```
<h1>
  Title
  <Icon />
</h1>
```


aigoncharov (approved):
LGTM!

packages/ui/src/Icon.tsx
```diff
@@ -5,19 +5,38 @@ import styleConsts from '../styles/constants/export.scss'
 
 type IconSize = 'small' | 'normal'
 
-export interface IconProps {
+// Makes it required to set either "aria-label", "aria-labelledby" or "aria-hidden" attribute.
+type AriaProps =
+  | {
```
jvorcak, line 10: nit: Maybe we could introduce this helper and have this more readable?

`type AtLeastOne<T, U = {[K in keyof T]: Pick<T, K> }> = Partial<T> & U[keyof U]`

https://stackoverflow.com/questions/48230773/how-to-create-a-partial-like-that-requires-a-single-property-to-be-set

aigoncharov: @jankoritak we also have to force that other properties are not passed, so it should be a bit different
```ts
type ExactlyOneOf<T> = { [K in keyof T]: Pick<T, K> & { [E in Exclude<keyof T, K>]?: never }}[keyof T]
```
https://www.typescriptlang.org/play?#code/C4TwDgpgBAYgrgGwQGQJYDsDWAlCBnYAZQHs4AnAY2gF4oBvAKCigFtiATCBALigLIwBzANxMoZLr35DRzYKmAIIU4APQiGAXwYNQkKAFEAHgEMKikAHl0ESwDMAPABUAfFFp0oAbQDSUDFCYECDEdlBOALq8AAqoFJjOADRQPm4AZPTeBv7ohkYUCHCcDkEhYU7JqREA-Lw2AG4QZFCaml6loeERuuDQaFi4BCTkVO55ZhbWto7wSP04+ESklBAuOhTE6ARQxJgmIAAMvPODSyM0mWycPFAA5Ca3yRI3twBGj1Dyisp3FLctDA2W2AOz2IAAjMcMAshstRh5WBxJHcHk9kW9-togdtdvsAExQgaLYYrMaeK7oh4A7Eg3EgADMhJhZ1JCOevAx1M2OLBABYmacSfDMl8lBy-gDAdyQU0yMQyEcoCdiXCLnRNFAAPSa1ioPB4IRQMByyBkUBS4FQWXyyFK6GC1VkxHXDmoz4KdESjXaq35fB4I0mpqgHa5aTqKAID1kEwIBhAA

However, it is really complex to digest. Should we stick to simpler stuff for readability?
On the other hand, it looks pretty cool! Thnx for finding it @jankoritak! I would suggest leaving a good comment with documentation on how to apply it and what it does if we decide to introduce it.

jvorcak (approved):

PR #69, opened 2020-11-11
(Feature) Card component

### LINKED ISSUE #11: [UI core library] Card
https://zeroheight.com/11d0e6dac/p/432e36-card

1. Create a  component with API
    1. content (as children)
    2. type
    3. title
    4. docs

jozefhruska:
Closes #11.


packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,47 @@
...
+      }
+    }
+  }
+
+  .content {
+    @include c.typographyBodyNormal;
+
+    padding: c.$grid * 3 c.$grid * 6 c.$grid * 6;
+
+    &.woTitle {
```
jvorcak, line 43: nit: What does this `wo` prefix stand for? `without`? I think it'd be more readable to use that.

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    if (iconButton !== undefined) {
+      return (
+        <IconButton kind="transparent" color={styleConsts.colorPrimary} size={type === 'primary' ? 'normal' : 'small'} {...iconButton} />
+      )
+    }
+
+    return null
+  }
```
aigoncharov, line 57: nit
This way we do not create a function just to return something from it right away in the same render. Less entities, less work for GC, right?
```suggestion
let icon: JSX.Element = null
if (icon !== undefined) {
  icon = <Icon icon={icon} color={styleConsts.colorPrimary} size={type === 'primary' ? 'normal' : 'small'} ariaHidden />
}
if (iconButton !== undefined) {
   icon = (
        <IconButton kind="transparent" color={styleConsts.colorPrimary} size={type === 'primary' ? 'normal' : 'small'} {...iconButton} />
   )
}
```

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { FC } from 'react'
+import { Icon, IconProps } from './Icon'
+import cn from 'classnames'
+
+import { IconButton, IconButtonProps } from './IconButton'
+
+import styleConsts from '../styles/constants/export.scss'
+import styles from './Card.module.scss'
+
+type CardHeaderProps =
```
aigoncharov, line 10: Great typings!

packages/ui/src/IconButton.tsx
```diff
@@ -14,9 +14,19 @@ export type IconButtonProps = {
   size?: IconProps['size']
   iconClassName?: string
   kind?: IconButtonKind
+  color?: IconProps['color']
 } & Pick<ButtonHTMLAttributes<HTMLButtonElement>, 'onClick' | 'className' | 'autoFocus' | 'disabled' | 'type'>
```
aigoncharov, line 18:
```suggestion
} & IconButtonKindProps & Pick<ButtonHTMLAttributes<HTMLButtonElement>, 'onClick' | 'className' | 'autoFocus' | 'disabled' | 'type'>

type IconButtonKindProps = { kind: 'primary', color?: never } | { kind?: 'transparent', color?: IconProps['color'] }
```

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,87 @@
...
+  | {
+      title?: never
+      iconButton?: never
+      icon?: never
+    }
+
+export type CardProps = {
+  type?: 'primary' | 'secondary' | 'highlighter'
+  icon?: IconProps['icon']
+} & CardHeaderProps
```
jvorcak, line 30: Should we use `DataTestProp` and allow to pass a custom data-test like we decided to do for TextField & Check/Radio buttons? So that the code base is consistent?

jvorcak (changes requested):
GJ Overall!

This is a bit out of scope of this PR implementation, but don't you think we could maybe increase contrast a bit? 
On some older screens with bad contrast this might not be visible at all (or when using an old projector in a meeting).


BTW - This is how it looks at IE11, but that's a different (imo low-prio) issue there might be some problem with box-shadow

Screenshot 2020-11-12 at 9 53 04

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,87 @@
...
+              {title}
+              {renderIcon()}
+            </h2>
+          ) : (
+            <h3 data-test="card-heading">
+              {title}
+              {renderIcon()}
+            </h3>
+          )}
+        </div>
```
aigoncharov, line 80: nit
```suggestion
          {type === 'primary' ? (
            <h2 data-test="card-heading" className={styles.header}>
              {title}
              {renderIcon()}
            </h2>
          ) : (
            <h3 data-test="card-heading" className={styles.header}>
              {title}
              {renderIcon()}
            </h3>
          )}
```

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,47 @@
...
+      align-items: center;
+      padding: 0;
+      margin: 0;
+
+      > svg,
+      > button {
+        margin-left: c.$grid * 2;
+      }
+    }
+  }
```
jankoritak, line 36: Nit: I would prefer having an explicit class on the `Icon`, resp. `IconButton`. The `svg`/`button` looks kinda out-of-place to me. It's a major nit though.

packages/ui/src/Card.module.scss
jankoritak: We can get rid of this class if we let `.content` handle the padding at all times:

Alter `.content`: 
from `padding: c.$grid * 3 c.$grid * 6 c.$grid * 6;` to `padding: c.$grid * 6;`

Alter `.header`:
from `padding: c.$grid * 6 c.$grid * 6 c.$grid * 3;` to `padding: c.$grid * 6 c.$grid * 6 0`

jankoritak (commented):
Great job overall! Leaving 2 suggestions.

packages/ui/src/Card.module.scss
jvorcak: Even better!

jankoritak: It actually "resolves" both this and my comment, we gotta keep this open. :)

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,47 @@
+@use '../styles/constants' as c;
+
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  align-content: stretch;
+  box-shadow: 0 c.$cardShadowSize c.$cardShadowBlur 0 c.$colorShadow;
```
jvorcak, line 7: Should this be called `c.$cardShaddowColor` instead of `c.colorShadow`?

jozefhruska: I just followed the convention for color names. We don't have any component-specific colors yet and I don't think we'll have multiple shadow colors. I'd rather keep this name, WDYT?

packages/ui/src/Card.module.scss
jozefhruska: @jankoritak I don't know why, but for some reason I can't reply to your comment :D Good idea, I'll apply your suggestion 👍 
image

jozefhruska:
@jvorcak I absolutely agree on the contrast issue. I implemented it according to design on Zeroheight, but we have different colors defined. Pawel told me he'll revisit Card component later.

packages/ui/src/Card.module.scss
jozefhruska: Agree, done 👍

packages/ui/src/Card.tsx
jozefhruska: Sure, didn't know that was a thing, added 👍

aigoncharov:
@jozefhruska visual-regression test suite failed. Does it pass for you locally?

jozefhruska:
@aigoncharov Sorry, fixed now 👍

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+      data-test={dataTest ?? 'card-wrapper'}
+      className={cn(styles.wrapper, {
+        [styles.typeSecondary]: type === 'secondary',
+        [styles.typeHighlighter]: type === 'highlighter',
+      })}
+    >
+      {title && (
+        <>
+          {type === 'primary' ? (
+            <h2 data-test="card-heading" className={styles.heading}>
```
aigoncharov, line 67: Should we include `typographyH2` for `h2`?

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+              {title}
+              <span className={styles.icon}>{iconElement}</span>
+            </h2>
+          ) : (
+            <h3 data-test="card-heading" className={styles.heading}>
```
aigoncharov, line 72: Should we include `typographyH3` for `h3`?

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+              {title}
+              <span className={styles.icon}>{iconElement}</span>
+            </h3>
+          )}
+        </>
+      )}
+
+      <div data-test="card-content" className={cn(styles.content)}>
```
aigoncharov, line 80:
```suggestion
      <div data-test="card-content" className={styles.content}>
```

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,39 @@
+@use '../styles/constants' as c;
+
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  align-content: stretch;
```
aigoncharov, line 6: Did you mean `align-items`?

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,39 @@
...
+  .heading {
+    display: flex;
+    align-items: center;
+    padding: c.$grid * 6 c.$grid * 6 0;
+    margin: 0;
+
+    .icon {
+      display: flex;
+      align-items: center;
+      margin-left: c.$grid * 2;
```
aigoncharov, line 30: nit: redundant styles
```suggestion
      margin-left: c.$grid * 2;
```

packages/ui/src/Card.module.scss
jozefhruska: Actually, that doesn't need to be there at all, thanks 🚀

packages/ui/src/Card.tsx
jozefhruska: Good catch, didn't realize in was global just for storybook preview 🙏

packages/ui/src/Card.tsx
jozefhruska: Updated, thanks 👍

packages/ui/src/Card.tsx
jozefhruska: Resolved together with `h3`.

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <>
+          {type === 'primary' ? (
+            <h2 data-test="card-heading" className={styles.heading}>
+              {title}
+              <span className={styles.icon}>{iconElement}</span>
+            </h2>
+          ) : (
+            <h3 data-test="card-heading" className={styles.heading}>
+              {title}
+              <span className={styles.icon}>{iconElement}</span>
```
danad02, line 74: nit: 
```suggestion
              {iconElement && <span className={styles.icon}>{iconElement}</span>}
```

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+        [styles.typeSecondary]: type === 'secondary',
+        [styles.typeHighlighter]: type === 'highlighter',
+      })}
+    >
+      {title && (
+        <>
+          {type === 'primary' ? (
+            <h2 data-test="card-heading" className={styles.heading}>
+              {title}
+              <span className={styles.icon}>{iconElement}</span>
```
danad02, line 69: nit:
```suggestion
             {iconElement && <span className={styles.icon}>{iconElement}</span>}
```

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+  | {
+      title?: never
+      iconButton?: never
+      icon?: never
+    }
+
+export type CardProps = {
+  type?: 'primary' | 'secondary' | 'highlighter'
+  icon?: IconProps['icon']
```
danad02, line 30: Should we remove `icon?: IconProps['icon']` here since it is part of `CardHeaderProps`?

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,45 @@
+@use '../styles/constants' as c;
+
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  box-shadow: 0 c.$cardShadowSize c.$cardShadowBlur 0 c.$colorShadow;
+  border-radius: c.$borderRadius;
+
+  &.typeSecondary {
+    background: none;
```
danad02, line 10: Since we're removing background here, should we add a `background-color: c.$colorNeutralWhite` to .wrapper? I am not sure

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,45 @@
...
+    border: c.$borderWidth solid c.$colorNeutralLight;
+    box-shadow: none;
+  }
+
+  &.typeHighlighter {
+    background: c.$colorInfoLight;
+    box-shadow: none;
```
danad02, line 17: Should we remove `box-shadow` here? What if it is a `primary` Card?

packages/ui/__tests__/Card.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    {
+      type: 'secondary',
+    },
+  ]
+
+  it.each(cardLayoutTestData)('Renders correct layout for specific component configurations.', async ({ headingType, ...rest }) => {
+    const wrapper = await mountAndCheckA11Y(<Card {...rest}>{cardContent}</Card>)
+
+    if (rest.title) {
+      expect(wrapper.findDataTest('card-heading').first().getDOMNode().tagName).toBe(headingType)
```
danad02, line 49: You can safely remove `.first()` here:
```suggestion
      expect(wrapper.findDataTest('card-heading').getDOMNode().tagName).toBe(headingType)
```

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+type CardHeaderProps =
+  | {
+      title: string
+      icon?: IconProps['icon']
+      iconButton?: never
+    }
+  | {
+      title: string
+      iconButton?: Omit<IconButtonProps, 'size' | 'kind'>
```
danad02, line 19: Should we rename it to `iconButtonProps`?

packages/ui/__tests__/Card.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+      expect(wrapper.findDataTest('card-heading').length).toBe(0)
+    }
+
+    if (rest.icon) {
+      expect(wrapper.find(Icon).length).toBe(1)
+    }
+
+    if (rest.iconButton) {
+      expect(wrapper.find(IconButton).length).toBe(1)
+    }
```
danad02, line 62: nit: We might be a little bit more specific:
```suggestion
   if (rest.icon) {
      expect(wrapper.find(Icon).props()).toEqual<IconProps>({
        icon: rest.icon,
        color: styleConsts.colorPrimary,
        size: rest.type === 'primary' ? 'normal' : 'small',
        ariaHidden: true,
      })
    }

    if (rest.iconButton) {
      expect(wrapper.find(IconButton).props()).toEqual<IconButtonProps>({
        kind: 'transparent',
        color: styleConsts.colorPrimary,
        size: rest.type === 'primary' ? 'normal' : 'small',
        ...rest.iconButton,
      })
    }
```

packages/ui/src/Card.tsx
danad02: It's literally props of `IconButton` component. When I see `iconButton` prop I instantly think about `IconButton` itself.
For `icon` we are picking a single prop which is a component so it makes sense to call it `icon` and not `iconProps`.

packages/ui/src/Card.module.scss
danad02: Ah I've overlooked that. Cool.

packages/ui/src/Card.module.scss
danad02: @jozefhruska But does `primary` have no background as well?

jozefhruska: Yes, you're right. Thanks 🙏 Sorry I didn't understand your comment at first 😌

packages/ui/src/Card.tsx
jozefhruska: Gues that makes sense, I'll update it. Thanks ❤️

danad02 (approved):
LGTM! :+1:

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,84 @@
...
+        [styles.typeSecondary]: type === 'secondary',
+        [styles.typeHighlighter]: type === 'highlighter',
+      })}
+    >
+      {title && (
+        <>
+          {type === 'primary' ? (
+            <h2 data-test="card-heading" className={styles.heading}>
+              {title}
+              {iconElement && <span className={styles.icon}>{iconElement}</span>}
```
danad02, line 68: @jozefhruska Is the icon supposed to be on the right? In figma example, it is on the left: https://www.figma.com/file/8mVm6LTbp2Z0RaWWjTZoft/%F0%9F%90%9DHIVE---Hazelcast-Design-System?node-id=10077%3A2

aigoncharov (approved):
Should be good to be merged once https://github.com/hazelcast/frontend-shared/pull/69#discussion_r528464789 is resolved

packages/ui/src/Card.tsx
jozefhruska: When I started there were only Zeroheight designs. We should probably review designs entirely, I'll ask Pawel.

aigoncharov:
@jozefhruska do you have an ETA for the task?

jozefhruska:
@aigoncharov Should be able to finish this before the end of this week (had to prioritize Cloud redesign over the past few days).

Issue #59, opened 2020-11-02
Consider moving shared dependencies from root to child packages


aigoncharov:
And use `lerna bootsrap --host` to lift them up

PR #72, opened 2020-11-13
Feature/select

### LINKED ISSUE #16: [UI core library] Select
https://zeroheight.com/11d0e6dac/p/12bbb4-dropdown

Implement only the classic dropdown menu initially

        1. name
        2. value
        3. onChange
        4. onBlur
        5. label
        6. error
        7. touched
        8. options (as children)
2. Create a formik v2 wrapper

jankoritak:
Closes #16


packages/ui/src/Select.tsx
```diff
@@ -0,0 +1,97 @@
...
+  errorClassName,
+  label,
+  name,
+  onBlur,
+  onChange,
+  options,
+  required,
+  selectClassName,
+  value,
+  notSelectedPlaceholder = 'Select',
```
aigoncharov, line 52: nit: I would suggest something language-neutral like `-`

packages/ui/src/Select.tsx
```diff
@@ -0,0 +1,97 @@
...
+          [styles.disabled]: disabled,
+          [styles.hasError]: error,
+          [styles.empty]: !value,
+        },
+        className,
+      )}
+    >
+      <Label id={idRef.current} label={label} />
+      <div className={styles.selectContainer}>
+        <select
```
aigoncharov, line 71: If we use native `select` than our `Select` is going to look quite differently compared to `Multiselect`. Should we use some library that can cover both cases for consistent look and feel?

aigoncharov: I think `react-select` worked out just perfectly for us at MC. I am researching other options now.

jvorcak: @aigoncharov I agree with using one lib for both scenarios

jankoritak: Good point! I totally overlooked the fact there will be a multi-select component. As much as I don't like `react-select` styling API, I think we should stick to it. It has proven to be a great match for us.

jvorcak: @aigoncharov @jankoritak we already use react-select as well on .cloud.

aigoncharov: @jankoritak @jankoritak there is an alternative to `react-select`. We could use low-level [downshift](https://github.com/downshift-js/downshift) and build a custom select ourselves. We already have `popper` to position the menu, so we'll just need to style it. However, I could imagine it might take a while to do it right.

aigoncharov: I gave it some thinking and I feel like we would be better off with `reatc-select`. It might be a bit bulky, but we have saved KB with other components, right? Time is of the matter, we should start integration ASAP. We could always reimplement the component matching the existing API later on when we have time. Let's do simple stuff for now. WDYT?

aigoncharov: @jankoritak could you prioritize this task over your other ones? I think `Select` and `Password` are only ones left to start with the new login flow for MC.

jankoritak: Certainly, I'll start with this right away!

danad02: Implementing the components(s) using downshift would take a lot of time. I would stick to `react-select`, at least for now.

package.json
```diff
@@ -12,6 +12,7 @@
     "@types/jest": "^26.0.14",
     "@types/react": "^16.9.51",
     "@types/react-dom": "^16.9.8",
+    "@types/react-select": "^3.0.26",
```
aigoncharov, line 15: We will move this types to `@hazelcast/ui` later in scope of https://github.com/hazelcast/frontend-shared/issues/59

packages/ui/src/SelectField.tsx
```diff
@@ -0,0 +1,145 @@
...
+    <div
+      data-test={dataTest}
+      className={cn(
+        styles.container,
+        {
+          [styles.disabled]: disabled,
+          [styles.hasError]: error,
+          [styles.empty]: !value,
+        },
+        // Menu container is either this select itself ot document.body
```
danad02, line 109:
```suggestion
        // Menu container is either this select itself or document.body
```

packages/ui/src/SelectField.tsx
```diff
@@ -0,0 +1,145 @@
...
+          className="hz-select-field"
+          classNamePrefix="hz-select-field"
+          // https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute
+          aria-errormessage={error && errorId(idRef.current)}
+          aria-invalid={!!error}
+          aria-required={required}
+          isClearable={isClearable ?? false}
+          isDisabled={disabled}
+          isMulti={false}
+          isSearchable={isSearchable}
```
danad02, line 129: I would love to see multi select and searchable select stories.

packages/ui/src/SelectField.tsx
```diff
@@ -0,0 +1,145 @@
...
+    <div
+      data-test={dataTest}
+      className={cn(
+        styles.container,
+        {
+          [styles.disabled]: disabled,
+          [styles.hasError]: error,
+          [styles.empty]: !value,
+        },
+        // Menu container is either this select itself or document.body
```
danad02, line 109: > Menu container is either this select itself or document.body
According to `getMenuContainer` it could be any element. Or am I mixing two different things here?

packages/ui/src/SelectField.module.scss
```diff
@@ -0,0 +1,141 @@
...
+    padding-right: c.$grid * 2;
+    box-sizing: border-box;
+    border: c.$borderWidth solid c.$colorNeutralLight;
+    border-radius: c.$borderRadius;
+    background-color: c.$colorNeutralLighter;
+    // Note: Is IE11 problematic in case of checkbox/radio, but not select
+    /* doiuse-disable css-appearance */
+    //appearance: none;
+    /* doiuse-enable css-appearance */
+    //width: 100%;
```
danad02, line 68: I can't quite get what's going on here from the message :cry: .

packages/ui/__tests__/SelectField.test.tsx
```diff
@@ -0,0 +1,214 @@
+import React from 'react'
+import { v4 as uuid } from 'uuid'
+// import { act } from 'react-dom/test-utils'
```
danad02, line 3:
```suggestion
```

packages/ui/__tests__/SelectFormik.test.tsx
```diff
@@ -0,0 +1,100 @@
+import React, { createRef } from 'react'
```
danad02, line 1: nit: Can we rename the file to `SelectFieldFormik.test.tsx` for consistency?

danad02 (commented):
@aigoncharov Left several comments.

packages/ui/src/SelectField.tsx
aigoncharov: Totally forgot about searchable select. Thnx. As to multiselect. I think it should be a different component.

aigoncharov: Addressed in https://github.com/hazelcast/frontend-shared/pull/72/commits/acbad02b1ebd440112f7c2ef53b08a90a6b49557

packages/ui/src/SelectField.tsx
```diff
@@ -0,0 +1,145 @@
...
+  label: string
+  required?: boolean
+  helperText?: string | ReactElement
+  className?: string
+  errorClassName?: string
+  placeholder?: string
+  menuPortalTarget?: 'body' | 'self' | HTMLElement | null
+} & DataTestProp &
+  Pick<InputHTMLAttributes<HTMLElement>, 'autoFocus' | 'disabled' | 'autoComplete'> &
+  Pick<ReactSelectProps, 'isSearchable' | 'menuIsOpen' | 'menuPlacement' | 'noOptionsMessage' | 'inputValue'>
```
jankoritak, line 58: Nit: We can pick `required` and `placeholder` from `InputHTMLAttributes` as well.

```suggestion
export type SelectFieldExtraProps<V> = {
  options: SelectFieldOption<V>[]
  label: string
  helperText?: string | ReactElement
  className?: string
  errorClassName?: string
  menuPortalTarget?: 'body' | 'self' | HTMLElement | null
} & DataTestProp &
  Pick<InputHTMLAttributes<HTMLElement>, 'autoFocus' | 'disabled' | 'autoComplete' | 'placeholder' | 'required'> &
  Pick<ReactSelectProps, 'isSearchable' | 'menuIsOpen' | 'menuPlacement' | 'noOptionsMessage' | 'inputValue'>
```

packages/ui/src/SelectField.tsx
danad02: @aigoncharov So we're going to use `SelectField` as a base for `MultiSelectField` component, right?

jankoritak (commented):
Stellar work! Cannot approve due to being author of the PR. Leaving that up to @danad02.

danad02 (approved):
Great job! Ready to merge once we settle on https://github.com/hazelcast/frontend-shared/pull/72#discussion_r535131045

packages/ui/src/SelectField.tsx
aigoncharov: @danad02 currently, `SelectField` does not accept `isMulti` as a prop, so you won't be able to do it unless you change the code :) I just feel like MultiSelect has a completely different look and feel, so I would rather do it as a standalone component.

danad02: Ah got it :smile: Well there is no information about it in the style guide. But if the future designs tell us to use this as a multiselect then there is nothing easier than enabling it and change styles a bit. Approved!

aigoncharov: @danad02 I think we will implement MultiSelect based on the same react-select. I just do not want to do with lots of conditional code to make it happen. I would rather copy-paste something from this component into a new shiny `MultiSelect` and make it works there for the multi select use-case only

danad02: Yeah there would probably be a surprising number of conditions :+1:

PR #75, opened 2020-11-17
Update RadioGroup.stories.tsx

jvorcak:


packages/ui/__stories__/RadioGroup.stories.tsx
```diff
@@ -48,7 +48,7 @@ export const RadioGroupWrappedInFormik = () => {
         name: 'Server Error: Invalid name',
       }}
       validate={(values) => {
-        const errors: Partial<Values> = {
+        const errors: Partial<{ [key in keyof Values]: string }> = {
```
aigoncharov, line 51:
```suggestion
        const errors: Partial<Values> = {
```

aigoncharov, line 51: Why? Isn't it the same thing?

jvorcak: @aigoncharov in this particular case yes. But if there are other types in values e.g. numbers, the old solution won't work. So I just want this to be consistent across all formik-related stories.

aigoncharov (approved):

jankoritak (approved):

PR #76, opened 2020-11-18
Adds formik related stories & modifies error property

jvorcak:
⚠️ The most important part of the PR: I think that we should not play around with `touched` already in the wrappers, it can cause problems e.g. with server side validation, if you want to see it try to put it back and have a look at the stories. AFAIK the initial error should be shown even if you have not touched the field.
✅ Added basic Formik stories for all form-related inputs
✅ Made placeholder for a number field optional  - we can discuss this, but I think only `name` and `label` should be required.


aigoncharov:
@jvorcak why do you think we should display the error if the input is not touched? 
The problem with displaying the errors for the untouched field is this. Imagine you have 10 fields. All of them are required. If we display the errors for the untouched fields, all the fields in the form will have the required error message displayed. AFAIK, displaying errors for fields user did not have a chance to touch yet is bad UX. 
I have never used the initialErrors myself, but I see your point. What if instead of removing `meta.touched` we enhance the condition?
I suggest something like:
```
meta.touched || meta.initialError === meta.error ? meta.error : undefined
```
BTW, I would suggest to use field-level validation in our stories as it is more popular way to validate fields, at least in MC team :) I would also suggest to add the `submit` button to the formik related stories.

jvorcak:
@aigoncharov let's say you do SSR, and the form is rendered with initialErrors that are given to the user by a server side validation. On the other hand I see the need for touched prop. I think your suggestion covers it all.

> BTW, I would suggest to use field-level validation in our stories as it is more popular way to validate fields, at least in MC team :) I would also suggest to add the submit button to the formik related stories.

That I agree with.

What about an optional placeholder, is it ok for it to be optional?

packages/ui/__tests__/RadioGroupFormik.test.tsx
```diff
@@ -140,6 +140,5 @@ describe('RadioGroupFormik', () => {
 
     wrapper.update()
     expect(wrapper.find('div').contains('Server Error: Invalid name')).toBeFalsy()
-    expect(wrapper.find('div').contains('Aragorn is stronger!')).toBeTruthy()
```
aigoncharov, line 143: WHy are we removing it?

packages/ui/__tests__/NumberFieldFormik.test.tsx
```diff
@@ -111,16 +113,17 @@ describe('NumberFieldFormik', () => {
...
-      simulateChange(wrapper.find('input'), '56')
+      simulateChange(wrapper.find('input'), 56)
     })
-    wrapper.update()
 
-    // The error is displayed only when the input becomes dirty
-    expect(wrapper.find(Error).prop('error')).toBe('Dark side')
+    expect(formikBag.current?.values).toEqual({
+      name: 56,
+    })
```
aigoncharov, line 126: nit: we are testing the error here. I would not check the values in this test, just the errors

aigoncharov:
@jvorcak optional placeholder makes total sense. Thnx!

packages/ui/__tests__/RadioGroupFormik.test.tsx
```diff
@@ -139,6 +139,16 @@ describe('RadioGroupFormik', () => {
...
     wrapper.update()
+
+    // let's click other valid option
+    // eslint-disable-next-line @typescript-eslint/require-await
+    await act(async () => {
+      wrapper.find("input[value='gandalf']").simulate('blur')
+    })
+
+    wrapper.update()
+
```
aigoncharov, line 150: nit
```suggestion
      wrapper.find("input[value='gandalf']").simulate('blur')
     })

    wrapper.update()

```

aigoncharov:
@jvorcak could you generate the screenshots?
Also should we add those submit buttons?
BTW, if we are using form-level validators, we should return `undefined` from them in case of no error. Otherwise, the form cannot be submitted.

packages/ui/__tests__/RadioGroupFormik.test.tsx
jvorcak: Does it work for you? For me it doesn't

packages/ui/src/NumberField.tsx
```diff
@@ -28,7 +28,7 @@ export type NumberFieldExtraProps = {
   className?: string
```
danad02, line 28: I found a bug in `NumberField` not related to this PR changes.
When you remove the value from the number field, you get this error: 
image

In the component we try to parse the input value with `parseInt` or `parseFloat`. However we do not deal with an empty string `''`. If you try to parse an empty string, you get `NaN`.

danad02: Also we should specify plus and minus `IconButton` prop `type='button'`, otherwise we're getting this warning:
image

packages/ui/__tests__/RadioGroupFormik.test.tsx
```diff
@@ -139,6 +139,15 @@ describe('RadioGroupFormik', () => {
     })
```
danad02, line 139: nit: I'd suggest to make the content of this look like this:
```
const wrapper = await mountAndCheckA11Y(<TestForm />)

    expect(wrapper.find(RadioGroup).prop('error')).toBe('Server Error: Invalid name')

    // let's click other valid option
    // eslint-disable-next-line @typescript-eslint/require-await
    await act(async () => {
      wrapper.find("input[value='gandalf']").simulate('change')
    })
    // eslint-disable-next-line @typescript-eslint/require-await
    await act(async () => {
      wrapper.find("input[value='gandalf']").simulate('blur')
    })
    wrapper.update()

    expect(wrapper.find(RadioGroup).prop('error')).toBe('Aragorn is stronger!')
```

packages/ui/__tests__/RadioGroupFormik.test.tsx
danad02: It doesn't work for me eaither.

packages/ui/__tests__/NumberFieldFormik.test.tsx
```diff
@@ -111,16 +113,14 @@ describe('NumberFieldFormik', () => {
 
     const wrapper = await mountAndCheckA11Y(<TestForm />)
 
-    expect(wrapper.find(Error).prop('error')).toBe(undefined)
+    expect(wrapper.find(Error).prop('error')).toBe('Dark side')
```
danad02, line 116: nit:
```suggestion
    expect(wrapper.find(NumberField).prop('error')).toBe('Dark side')
```

packages/ui/__tests__/NumberFieldFormik.test.tsx
```diff
@@ -111,16 +113,14 @@ describe('NumberFieldFormik', () => {
...
     // eslint-disable-next-line @typescript-eslint/require-await
     await act(async () => {
-      simulateChange(wrapper.find('input'), '56')
+      simulateChange(wrapper.find('input'), 56)
     })
-    wrapper.update()
 
-    // The error is displayed only when the input becomes dirty
-    expect(wrapper.find(Error).prop('error')).toBe('Dark side')
+    expect(wrapper.update().find(Error).prop('error')).toBe('Client side validation')
```
danad02, line 124: nit: 
```suggestion
    expect(wrapper.update().find(NumberField).prop('error')).toBe('Client side validation')
```

danad02 (commented):
@jvorcak Left few nits and one bug report. The bug is not related to this PR but it would be great to fix it ASAP.

packages/ui/src/NumberField.tsx
danad02: Fix in https://github.com/hazelcast/frontend-shared/pull/82

aigoncharov:
@jvorcak could you finish the PR?

jvorcak:
@aigoncharov sorry for a delay, I'll try to do it today/latest by tomorrow morning.

aigoncharov (approved):

PR #79, opened 2020-11-19
Feature/slider

jvorcak:
Some notes, things to discuss:

- ✅ IE11 works, quite a lot of design decisions had to be done because of it :/
- Error is only visible through the message, no border is set on a thumb (because it'd be weird in a range mode). I can modify it it we have some designs for it
- ⚠️ no Designs are present in Figma yet, so there is a missing Storybook link to it
- We should follow https://www.w3.org/TR/wai-aria-practices-1.2/#slidertwothumb guidelines - basically we use 2 sliders on top of each other
- 🕐 I'll add screenshots in the end once the PR is approved to speed up the process.

What I'd need help with / possible improvements: 
- TS typeguards - maybe this could be done in a more elegant way, but I want a component to check that `value` & `onChange` are consistent with the parameter, that I could only done with generics so I combined it with typeguards
- Formik stories will be modified the same way as I'll do here https://github.com/hazelcast/frontend-shared/pull/76

Screenshot 2020-11-19 at 13 36 55


packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+type SingleValueChangeFn = (val: number, e?: ChangeEventType) => void
+type MultiValueChangeFn = (val: [number, number], e?: ChangeEventType) => void
+
+export type SliderCoreProps<T extends Value> = {
+  value: T
+  name: string
+  onChange: T extends number ? SingleValueChangeFn : MultiValueChangeFn
+  error?: string
+}
```
jozefhruska, line 29: Is this necessary?

```suggestion
export type SliderValueProps =
  | {
      value: number
      onChange: SingleValueChangeFn
    }
  | {
      value: [number, number]
      onChange: MultiValueChangeFn
    }

export type SliderCoreProps = {
  name: string
  error?: string
} & SliderValueProps
```

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,333 @@
...
+ * @param markValue - value of the mark
+ * @param max - max value of a slider
+ * @param firstValue - current position of the first slider
+ * @param secondValue - current position of the second slider, max if not in range mode
+ * @param isRange - determines whether we're in a range mode
+ * @return {
+ *   left: left position of the element in %
+ *   isActive: true if marker is in the filled area of the slider
+ * }
+ */
```
jozefhruska, line 66: Love the documentation 🚀

jozefhruska (commented):
Looks really cool, found just some minor things, mostly nitpicks! 🚀

packages/ui/src/Slider.module.scss
```diff
@@ -0,0 +1,261 @@
...
+  width: $thumbWidth;
+  height: $thumbWidth;
+}
+
+@mixin gridParent($cols: 1fr, $rows: 1fr) {
+  display: grid;
+  grid-template-columns: $cols;
+  grid-template-rows: $rows;
+  -ms-grid-columns: $cols;
+  -ms-grid-rows: $rows;
```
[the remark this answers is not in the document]
jvorcak: It should, but it doesn't, do we have it configured correctly? I decided to configure it this way for now and we can fix autoprefixer as part of this or other PR. I'll have a look soon.

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,343 @@
...
+    left: (markValue / max) * 100,
+    isActive: isRange ? markValue >= firstValue && markValue <= secondValue : markValue < firstValue,
+  }
+}
+
+/**
+ * Slider component
+ * https://www.w3.org/TR/wai-aria-practices-1.2/#slidertwothumb
+ */
+export function Slider<T extends Value = number>({
```
danad02, line 93: @jvorcak I think it would be great if we could see the current slider value(s), as part of the component. Can you ask Pawel about this?

packages/ui/src/Slider.module.scss
```diff
@@ -0,0 +1,261 @@
...
+  background: c.$colorPrimaryLight; // background will represent the dot inside
+  border: c.$sliderThumbBorderSize solid c.$colorNeutralWhite; // the white part will be done by the border
+  box-sizing: content-box; /* and because of all of this, we need content-box */
+  border-radius: 50%;
+  pointer-events: auto; /* catch clicks */
+  width: $thumbWidth;
+  height: $thumbWidth;
+}
+
+@mixin gridParent($cols: 1fr, $rows: 1fr) {
```
aigoncharov, line 33: I would suggest not to use grids because of IE11. Yes, they are supported there, but it an older spec, there are plenty of bugs. I just feel like it could seriously backfire. Isn't there a way to do the same styling without grids?
@danad02 @jankoritak @jvorcak @jozefhruska  WDYT?

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,343 @@
...
+    },
+    [firstValue, updateValues],
+  )
+
+  const width: number = ((secondValue - firstValue) / max) * 100
+  const left: number = firstValue / max
+
+  return (
+    <div className={className} data-test={dataTest}>
+      <label id={idRef.current} className={styles.label}>
```
aigoncharov, line 228: Should we use our own `Label`? I understand that we might have two inputs, so it is unclear, which one it should link to. However, the first input is always present, so we could link it to the first one.

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,343 @@
...
+              const { left } = getMarkMetadata(markValue, max, [firstValue, secondValue], isRange)
+              return (
+                <li key={markValue} style={{ left: `${left}%` }}>
+                  {label}
+                </li>
+              )
+            })}
+          </ul>
+        )}
+        {helperText && <Help parentId={idRef.current} helperText={helperText} className={styles.helperText} />}
```
aigoncharov, line 338: Last label seems kind of shifted
image

aigoncharov: Should we make the first and the last labels aligned with the slider left and right borders?

packages/ui/src/SliderFormik.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { FieldValidator, useField } from 'formik'
+import { Slider, SliderExtraProps } from './Slider'
+
+export type SliderFormikProps<V extends object> = SliderExtraProps & {
+  name: keyof V
+  validate?: FieldValidator
+}
+
+export const SliderFormik = <V extends object>({ name, validate, ...props }: SliderFormikProps<V>): ReactElement => {
+  const [field, meta, { setValue, setTouched }] = useField<number | [number, number]>({
```
aigoncharov, line 11: What if we add an explicit `range?: boolean` prop to the `Slider`? Then we would be able to strictly type the value that we get here or the one that we get in the `onChange` callback.

```tsx
type Value = typeof range extends true ? [number,number] : number

useField<Value>(..)

onChange = useCallback((newValue: Value) => ...)
```

packages/ui/src/SliderFormik.tsx
```diff
@@ -0,0 +1,33 @@
...
+    [setValue, setTouched],
+  )
+
+  return (
+    <Slider
+      {...props}
+      name={name}
+      value={field.value}
+      onChange={onChange}
+      error={meta.touched || meta.initialError === meta.error ? meta.error : undefined}
```
aigoncharov, line 30: Should we use `getFieldError` here?

packages/ui/__stories__/Slider.stories.tsx
```diff
@@ -0,0 +1,235 @@
...
+    </Formik>
+  )
+}
+
+/**
+ * The following stories are here mostly mostly for visual regression testing and do not demonstrate the way
+ * how this component should be used.
+ * @constructor
+ */
+export const DoNotUse__OpacityThumb__Beginning = () => {
```
aigoncharov, line 194: What exactly do they test? I do not quite get it

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,343 @@
...
+  className,
+  sliderClassName,
+  error,
+  helperText,
+  marks,
+  label,
+  disabled,
+  'data-test': dataTest,
+}: SliderProps<T>) {
+  /**
```
aigoncharov, line 109: Great comments!

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,343 @@
...
+          <div
+            className={styles.fill}
+            data-test="fill"
+            // Since there is no easy way hot to pass properties to SCSS, we need to use
+            // inline styles.
+            style={
+              isRange
+                ? {
+                    width: `${width}%`,
+                    marginLeft: `calc(${left * 100}%)`,
```
aigoncharov, line 296: Do we need calc here at all? Can't it be `${left * 100}%`?

packages/ui/src/Slider.module.scss
jvorcak: We can do absolute positioning, that should be quite easy.

packages/ui/src/Slider.tsx
jvorcak: According to https://www.w3.org/TR/wai-aria-practices-1.2/#slidertwothumb if we have 2 sliders we should reference them through `aria-label` or `aria-labelledby`. This is very special, but we should follow the spec and our current Label component doesn't have this API.

packages/ui/src/Slider.tsx
jvorcak: I've discussed this with @pawelklasa, but I'm open to change it if he says it should look differently.

packages/ui/src/SliderFormik.tsx
jvorcak: @aigoncharov of course we should, at the time of this PR it didn't exists, now it's definitely a way to go! Thanks for creating it.

packages/ui/src/Slider.tsx
aigoncharov: Can't we do both? I do not see any contradiction there? Quite the contrary, `label-for` does not prevent us from using `aria-labelledby`, but `label-for` adds easy navigation. Once user clicks on the label, they will be focused on the associated slider thumb. WDYT?

packages/ui/__stories__/Slider.stories.tsx
jvorcak: When implementing a slider there are these issues that I run into https://css-tricks.com/multi-thumb-sliders-particular-two-thumb-case/#issues

These screenshots test that whoever touches this component has it done correctly, one needs to try to combine thumb height and other variables and it still work correctly. I'll add some more comments there to explain.

packages/ui/src/Slider.module.scss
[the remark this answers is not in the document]
jvorcak: Since we will replace grid with absolute positioning, this is no longer relevant.

packages/ui/__stories__/Slider.stories.tsx
aigoncharov: Could you leave a comment with the link?

jvorcak:
@aigoncharov @jozefhruska thanks for a great feedback, I'll wrap this up during next week as I have a lot of job on .cloud right now.

packages/ui/src/Slider.tsx
jvorcak: Do you mean `htmlFor`? I'm not aware of `label-for`. 

Since there can be only one element with an unique ID, the only solution I see here is to pass this ID to the first slider always. And since the second one doesn't have a visible label, put `aria-label` there. That should follow the spec and fulfill your requirement of using our `Label` thus making clicking on label working.

jvorcak:
Ready for re-review, however we're still waiting for design inputs, especially for https://github.com/hazelcast/frontend-shared/pull/79#discussion_r530348593

packages/ui/src/Slider.tsx
pawelklasa: It's all good, I think the only slight addition might be making the selection label bold

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,340 @@
...
+      const { offsetX, target } = e
+      if (disabled) {
+        return false
+      }
+      if (!wrapperRef.current || wrapperRef.current !== target) {
+        return
+      }
+      const clickPercentage = offsetX / wrapperRef.current.offsetWidth
+      const resultStepSpot: number = Math.round((clickPercentage * max) / step) * step
+      if (!isRange || Math.abs(firstValue - resultStepSpot) < Math.abs(secondValue - resultStepSpot)) {
```
danad02, line 169: Love the distance checking for both handles for range slider!

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,340 @@
...
+    },
+    [firstValue, secondValue, onChange, value],
+  )
+
+  /**
+   * The following code makes sure that the relevant thumb is moved to the position once
+   * user clicks on a track.
+   *
+   * 1. We always move the first thumb in a non-range mode.
+   * 2. Otherwise, we move the thumb that is closed to a click event.
```
danad02, line 155:
```suggestion
   * 2. Otherwise, we move the thumb that is closest to the clicked point.
```

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+  useEffect(() => {
+    window.addEventListener('click', clickOnTrackHandler)
+    return () => {
+      window.removeEventListener('click', clickOnTrackHandler)
+    }
+  }, [clickOnTrackHandler])
+
+  /**
+   * The following callbacks reacts to native HTML events and update the values
```
danad02, line 189:
```suggestion
   * The following callbacks react to native HTML events and update the values
```

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const setFirstValueCallback = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const {
+        target: { value },
+      } = e
+      const newValue = parseInt(value, 10)
+      if (newValue > secondValue) {
+        return false
+      }
+      updateValues([newValue, null], e)
```
danad02, line 200: nit: a bit more readable imo.
```suggestion
      if (newValue <= secondValue) {
        updateValues([newValue, null], e)
      }
```

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const setSecondValueCallback = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const {
+        target: { value },
+      } = e
+      const newValue = parseInt(value, 10)
+      if (newValue < firstValue) {
+        return false
+      }
+      updateValues([null, newValue], e)
```
danad02, line 214: nit: a bit more readable imo.
```suggestion
      if (newValue >= firstValue) {
        updateValues([null, newValue], e)
      }
```

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,340 @@
...
+    },
+    [wrapperRef, firstValue, secondValue, disabled, isRange, max, step, updateValues],
+  )
+
+  useEffect(() => {
+    window.addEventListener('click', clickOnTrackHandler)
+    return () => {
+      window.removeEventListener('click', clickOnTrackHandler)
+    }
+  }, [clickOnTrackHandler])
```
danad02, line 186: We can use a hook from `react-use`, I tested it and it works as expected:
` useEvent('click', clickOnTrackHandler)`

packages/ui/src/SliderFormik.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const SliderFormik = <V extends object>({ name, validate, ...props }: SliderFormikProps<V>): ReactElement => {
+  const [field, meta, { setValue, setTouched }] = useField<number | [number, number]>({
+    name,
+    validate,
+  })
+
+  const onChange = useCallback(
+    (newValue: number | [number, number]) => {
+      setValue(newValue)
+      setTouched(true)
```
danad02, line 20: `setTouched` runs validation with **previous** value here. From documentation:

> Set touched imperatively. Calling this will trigger validation to run if validateOnBlur is set to true (which it is by default). You can also explicitly prevent/skip validation by passing a second argument as false.

So I suggest following:
```suggestion
      setTouched(true, false)
```

packages/ui/__tests__/SliderFormik.test.tsx
```diff
@@ -0,0 +1,88 @@
+import React, { createRef } from 'react'
+import { Form, Formik, FormikProps } from 'formik'
+import { mountAndCheckA11Y, simulateChange } from '@hazelcast/test-helpers'
+import { act } from 'react-dom/test-utils'
+
+import { SliderFormik } from '../src/SliderFormik'
+
+describe('SliderFormik', () => {
```
danad02, line 8: Can we also add tests for range slider?

packages/ui/__stories__/Slider.stories.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * The following stories are here mostly mostly for visual regression testing and do not demonstrate the way
+ * how this component should be used.
+ *
+ * The main goals of these stories is to make sure that the blue center of a thumb is exactly aligned with the
+ * track's border and the white border will just overflow as requested by a design.
+ *
+ * You can read about common implementation problems here
+ * https://css-tricks.com/multi-thumb-sliders-particular-two-thumb-case/#issues
```
danad02, line 197: That's useful to know! Thanks for adding the comment.

danad02 (changes requested):
@jvorcak Left several comments. Great job overall!

jvorcak:
Thanks @danad02 for the useful comments, I'll fix them once I get designs for showing the value.

packages/ui/src/SliderFormik.tsx
aigoncharov: I think this https://github.com/hazelcast/frontend-shared/pull/115 should resolve it. WDYT?

danad02: @aigoncharov Yes, I didn't notice it in other places, nice catch ;)

packages/ui/src/SliderFormik.tsx
jvorcak: @aigoncharov as we discussed on Slack last time, we will keep the current implementation with typeguards.

aigoncharov:
@jvorcak shouldn't it highlight the top label in this case and hide the bottom one?
image

jvorcak:
@aigoncharov good catch, I'll implement now.

packages/ui/__tests__/Slider.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    const onChange = jest.fn()
+    const wrapper = await mountAndCheckA11Y(
+      <Slider name="ram" label="RAM" value={[4, 10]} min={1} max={100} onChange={onChange} formatCurrentValue={(x) => `${x} RAM`} />,
+    )
+    expect(wrapper.findDataTest('slider-first-value-indicator').text()).toEqual('4 RAM')
+    expect(wrapper.findDataTest('slider-second-value-indicator').text()).toEqual('10 RAM')
+  })
+})
+
+describe('Slider helper functions', () => {
```
aigoncharov, line 226: Great idea! Totally agree that it is easier to test smaller units!

packages/ui/src/Slider.tsx
```diff
@@ -0,0 +1,374 @@
...
+
+  // Slider operates in two modes, either we have a single value slider or range value slider
+  const isRange: boolean = isRangeGuard(value)
+  const [firstValue, secondValue]: [number, number] = isRange ? (value as [number, number]) : [value as number, Infinity]
+
+  // wrapper ref is used to catch clicks on a track in order to move the closes thumb to the position
+  const wrapperRef = useRef<HTMLDivElement>(null)
+  const firstRangeInputRef = useRef<HTMLInputElement>(null)
+  const secondRangeInputRef = useRef<HTMLInputElement>(null)
+  const markValues: Array<number> = marks?.map(({ value }) => value) ?? []
```
aigoncharov, line 146: nit: Should we make it a `Set` for `O(1)`? Should we memoize it with `useMemo`?

aigoncharov (approved):
Great job! Left one last nit https://github.com/hazelcast/frontend-shared/pull/79/files#r541030562

danad02 (approved):
Cool. Very nice 👍

PR #82, opened 2020-11-23
Fix empty string handling for the `NumberField`

aigoncharov:
Closes the bug reported at https://github.com/hazelcast/frontend-shared/pull/76#discussion_r527585859


danad02 (approved):
Thanks for quick bug fix!

PR #84, opened 2020-11-23
export.scss -> export.module.scss

jvorcak:
Even though we now transpile our `@hazelcast/ui`, next.js still forbids to import files with `.css` / `.scss` extension unless they're in `_app.js/ts`

https://github.com/vercel/next.js/blob/master/errors/css-global.md 

This solution works, I've tried with `npm link` so far. @aigoncharov can we please release a new version of a lib to npm once this is merged?

----

An alternative to this would be to transpile, otherwise we will need to follow this simple rule.


aigoncharov (approved):
I do not see any potential issues with this. Good job!

PR #85, opened 2020-11-24
Feature/password field

### LINKED ISSUE #77: [Core UI] Password field

aigoncharov:
Closes #77


jvorcak:
Can we please add more string types for `TextField` as part of this PR, since we're touching the `TextField` in this PR?

https://www.w3schools.com/tags/att_input_type.asp

I think `email`, `search`, `url`, `tel` are at least relevant (maybe even hidden) and we will need them on .cloud.

aigoncharov:
> Can we please add more string types for `TextField` as part of this PR, since we're touching the `TextField` in this PR? (we can do it separately as well)
> 
> https://www.w3schools.com/tags/att_input_type.asp
> 
> I think `email`, `search`, `url`, `tel` are at least relevant (maybe even hidden) and we will need them on .cloud.

I suggest we do it in scope of a separate PR since all those types will need stories accordingly

packages/ui/src/PasswordField.module.scss
```diff
@@ -0,0 +1,20 @@
...
+  > input {
+    padding-right: 0;
+    margin-right: $inputMargin;
+  }
+
+  .toggle {
+    position: absolute;
+    right: $buttonMargin;
+    top: 50%;
+    transform: translateY(-50%);
```
jvorcak, line 18: nit: I'm thinking whether it wouldn't be more readable to use `margin: auto, top: 0, bottom: 0` and specify an explicit height instead of top50% and translateY.

packages/ui/src/utils/formik.ts
```diff
@@ -0,0 +1,3 @@
+import { FieldMetaProps } from 'formik'
+
+export const getFieldError = <T>(meta: FieldMetaProps<T>) => (meta.touched || meta.initialError === meta.error ? meta.error : undefined)
```
jvorcak, line 3: 👏 for this abstraction.

packages/ui/__stories__/PasswordField.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        return errors
+      }}
+      onSubmit={(values) => logger.log('submit', values)}
+    >
+      {({ values }) => (
+        <Form>
+          Values: {JSON.stringify(values)}
+          <PasswordFieldFormik<Values> name="password" label="Name" />
```
jvorcak, line 138: Since the other PR is already merged, please let's add a submit button and field level validation for consistency.

jvorcak (changes requested):
GJ 👏

jvorcak:
One thing I forgot about is missing export in `index.ts`, but that needs to be done after updating.

packages/ui/src/PasswordField.module.scss
aigoncharov: Don't you think that transform-translate is a more or less standard way to centralize an absolutely positioned element? Moreover, it would work with any input height.

jvorcak: Both are fine, thus => `nit`, maybe we can use grid for these kind of things? It's supported in IE11 with prefixes and solves a lot of these. WDYT?

aigoncharov: caniuse says IE11 supports only the older version of the spec https://caniuse.com/css-grid
I am kind of concerned with all possible issues. Have you tried using it for IE11 yourself?

jvorcak (approved):

packages/ui/src/PasswordField.module.scss
jvorcak: It is indeed limited and there is not a full support and only with prefixes, but for the basic usage like making a simple grid and position elements, I think it's totally doable also for production. I remember we used it for enterprise product while IE11 was required there and there were no complaints at all. But we were not overusing it that is true :)

danad02 (approved):

PR #89, opened 2020-11-24
Exports all current components

jvorcak:
This part was missing. Please make sure to add your stuff here once you implement something + check for this inside PRs.

I've renamed IconDescriptor and made it local for each component. Alternative is to create some common type, but that can be done out of the scope of this PR.


packages/ui/src/index.ts
```diff
@@ -1,7 +1,25 @@
+export * from './Alert'
+export * from './Badge'
+export * from './Button'
+export * from './Checkbox'
+export * from './CheckboxFormik'
 export * from './Error'
 export * from './Help'
 export * from './Icon'
+export * from './IconButton'
+export * from './Label'
```
aigoncharov, line 10: I would suggest leaving `Label` private

jvorcak: @aigoncharov let's say we will implement project specific components, this can be handy. 

e.g. maybe we will need this for card input (we will use Stripe's one) & connect this label, I think we could use this.

aigoncharov (approved):

danad02 (approved):

PR #95, opened 2020-11-25
Use useIsomorphicLayoutEffect

aigoncharov:
useIsomorphicLayoutEffect is still required in NumberField because onChange synchronously queues a state update in parent component and violates encapsulation and React warns - https://fb.me/setstate-in-render


jvorcak (approved):
💪

PR #108, opened 2020-12-03
Fixes Checkbox formik implementation

jvorcak:
- Fixes the way CheckboxFormik sets the value
- Added a unit test for this
- Added a story for this
- `CheckboxFieldFormik` renamed to `CheckboxFormik` for consistency


packages/ui/__tests__/CheckboxFormik.test.tsx
```diff
@@ -43,5 +43,40 @@ describe('CheckboxFormik', () => {
       },
       expect.anything(),
     )
+    expect(wrapper.find('input').props()).toHaveProperty('checked')
+  })
+
+  it('Checkbox is changed in a formik scenario', async () => {
```
aigoncharov, line 49: nit: SHould we combine it with a previous test like we do for some other components?

aigoncharov (approved):

packages/ui/__tests__/CheckboxFormik.test.tsx
jvorcak: @aigoncharov Since it's nit, I'll do that in a separate PR if it's OK, we'd like to get this merged soon, because we need it on .cloud

PR #109, opened 2020-12-04
Fix SelectField rendering in SSR env

### LINKED ISSUE #107: [UI core library] Select isn't SSR compatible
It seems like there's an issue with Select on SSR:

This is caused by this method (document is not defined during server side render):
```ts
const getMenuContainer = (menuPortalTarget: 'body' | 'self' | HTMLElement | null): HTMLElement | null | undefined => {
  if (menuPortalTarget == 'body') {
    return document.body
  }
  if (menuPortalTarget == 'self') {
    return undefined
  }
  return menuPortalTarget
}
```

Currently this is a blocker for the new Cloud sign up.

aigoncharov:
Closes #107

### DIFF
--- a/packages/ui/src/SelectField.tsx
+++ b/packages/ui/src/SelectField.tsx
@@ -13,6 +13,7 @@ import { Icon } from './Icon'
 import { IconButton } from './IconButton'
 
 import styles from './SelectField.module.scss'
+import { canUseDOM } from './utils/ssr'
 
 const DropdownIndicator = () => <Icon icon={ChevronDown} ariaHidden size="normal" className={styles.chevron} />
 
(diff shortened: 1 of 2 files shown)


danad02 (approved):

jvorcak (commented):
Late to the party! Good job on this one!

I'm thinking whether there is some method to automatically test on a unit-test level that components can be server-rendered. If you guys are aware of something like that, please share it with me :) Otherwise I'll try to figure it out in my free time.

aigoncharov:
> Late to the party! Good job on this one!
> 
> I'm thinking whether there is some method to automatically test on a unit-test level that components can be server-rendered. If you guys are aware of something like that, please share it with me :) Otherwise I'll try to figure it out in my free time.

@jvorcak I have the same thought. Will create a task for it.

PR #116, opened 2020-12-08
Add uuid support to SSR

### LINKED ISSUE #86: Replace useRef(uuid()) with react-uid
Problem when using our lib with SSR:

Screenshot 2020-11-24 at 12 02 15

Basically `useRef(uuid())` generates different string on a server compared to the client.

One possible solution:

Use: https://github.com/thearnica/react-uid
Related article: https://medium.com/@antonkorzunov/an-easy-way-to-name-labels-and-generate-unique-keys-5ec2f9667e6a

API would look like this

```
const uuid = useUID();

<label htmlFor={uuid} .... />
```

And all SSR rendered apps or tests would need to wrap the app with `<UIDReset>` from this lib. Client side only apps would need to do probably nothing.

https://github.com/thearnica/react-uid/blob/master/src/Control.tsx#L10

I've had a look at the source code of the lib and it seems reasonable, it's ~900B big. WDYT?

I'm open to other suggestions as well. cc @aigoncharov @jankoritak @jozefhruska @danad02

aigoncharov:
Closes #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 - [Installation](#installation)
   - [Under the hood](#under-the-hood)
 - [Usage](#usage)
+- [SSR](#ssr)
+  - [Stable IDs](#stable-ids)
 - [Project structure](#project-structure)
   - [Automated visual regression testing](#automated-visual-regression-testing)
 - [How-to](#how-to)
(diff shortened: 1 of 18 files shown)


danad02 (approved):
LGTM! :+1:

jvorcak:
LGMT

nit: It'd probably be nice to add a note about a need for using `UIDReset` for SSR rendered apps using this lib. 
https://github.com/thearnica/react-uid/blob/master/src/Control.tsx#L10

jvorcak (approved):

PR #117, opened 2020-12-08
Fix Chrome autofill

### LINKED ISSUE #113: Adjust margins of TextField when using the icon
This looks quite weird then Google password manager/autocomplete is used, which is quite common. The only thing we need is to adjust paddings/margins the same way as we do for password component.
Screenshot 2020-12-06 at 11 47 54
Screenshot 2020-12-06 at 11 47 42

Screenshot 2020-12-06 at 11 47 51
Screenshot 2020-12-06 at 11 47 40

aigoncharov:
Closes #113


danad02 (approved):
Nice stories to check the described behavior! :+1:

jvorcak (approved):
I agree with @danad02! GJ

PR #119, opened 2020-12-10
Card component

### LINKED ISSUE #11: [UI core library] Card

jozefhruska:
Closes #11


packages/ui/__stories__/Card.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const CardWithIconAndSeparator = () => (
+  <Card title="Card title" icon={Database} separator>
+    {cardContent}
+  </Card>
+)
+
+export const CardWithSmallText = () => (
+  <Card title="Card title">
+    <div className={styles.typographyBodySmall}>{cardContent}</div>
```
danad02, line 59: I don't think you can use mixins like this. There is no class on div:
image

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,37 @@
...
+    {title && (
+      <>
+        <h3 data-test="card-heading" className={styles.heading}>
+          {icon && <Icon icon={icon} className={styles.icon} ariaHidden />}
+          {title}
+        </h3>
+      </>
+    )}
+
+    {separator && <span data-test="card-separator" className={styles.separator} />}
```
danad02, line 31: In `Card.module.scss` we're using `display: block;` - can't we just change this element to `div` and remove the `display: block;`?

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,36 @@
+@use '../styles/constants' as c;
+
+.wrapper {
+  padding: c.$grid * 5;
+  box-shadow: 0 c.$cardShadowSize c.$cardShadowBlur 0 c.$colorShadow;
+  border-radius: c.$borderRadius;
+  background: c.$colorNeutralWhite;
+
+  h3.heading {
```
danad02, line 9: nit: I don't think we need to specify `h3` here:
```suggestion
  .heading {
```

packages/ui/__stories__/Card.stories.tsx
jozefhruska: Good catch! I wasn't trying to use the mixin, I just made a mistake while re-creating this PR. Thanks 🙏

packages/ui/src/Card.tsx
jozefhruska: Well, I think we can, but will it improve anything? It's just a habit of mine to use `span` elements as separators.

jozefhruska: @danad02 Maybe I could replace it with `<hr />`?

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,36 @@
...
+    @include c.typographyH3();
+
+    display: flex;
+    align-items: center;
+    padding: 0;
+    margin: 0 0 c.$grid * 4;
+
+    .icon {
+      line-height: 1;
```
danad02, line 18: nit: I removed the `line-height: 1;` and nothing changed - did you encounter a problem without it?

jozefhruska: Good catch, this is a leftover from the old Card design 🙏

packages/ui/src/Card.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  box-shadow: 0 c.$cardShadowSize c.$cardShadowBlur 0 c.$colorShadow;
+  border-radius: c.$borderRadius;
+  background: c.$colorNeutralWhite;
+
+  h3.heading {
+    @include c.typographyH3();
+
+    display: flex;
+    align-items: center;
+    padding: 0;
```
danad02, line 14: Is there a default padding on heading elements?

packages/ui/src/Card.tsx
danad02: @jozefhruska I take it back, `<hr />` should be used to represent a break between paragraphs.  

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr:
> Historically, this has been presented as a horizontal rule or line. While it may still be displayed as a horizontal rule in visual browsers, this element is now defined in semantic terms, rather than presentational terms, so if you wish to draw a horizontal line, you should do so using appropriate CSS.

I think we should use the `div` with styles as we have now.

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,37 @@
...
+ */
+export const Card: FC<CardProps> = ({ title, icon, separator = false, 'data-test': dataTest, children }) => (
+  <div data-test={dataTest ?? 'card-wrapper'} className={styles.wrapper}>
+    {title && (
+      <>
+        <h3 data-test="card-heading" className={styles.heading}>
+          {icon && <Icon icon={icon} className={styles.icon} ariaHidden />}
+          {title}
+        </h3>
+      </>
```
danad02, line 28:
```suggestion
      <h3 data-test="card-heading" className={styles.heading}>
        {icon && <Icon icon={icon} className={styles.icon} ariaHidden />}
        {title}
      </h3>
```

packages/ui/__tests__/Card.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    const icon = wrapper.find(Icon)
+    const title = wrapper.findDataTest('card-heading')
+    const separator = wrapper.findDataTest('card-separator')
+
+    if (props.title) {
+      expect(title.text()).toEqual(props.title)
+      expect(title.getDOMNode().tagName).toEqual('H3')
+    } else {
+      expect(title.length).toEqual(0)
```
danad02, line 42: nit: Personally, I find using `exists()` - `expect(title.exists()).toBe(false)` - to be a bit more readable :)

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,37 @@
...
+    )}
```
danad02, line 29:
```suggestion
    <h3 data-test="card-heading" className={styles.heading}>
      {icon && <Icon icon={icon} className={styles.icon} ariaHidden />}
      {title}
    </h3>
```

packages/ui/__tests__/Card.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    {
+      title: cardTitle,
+      icon: Database,
+      separator: true,
+    },
+    {
+      title: cardTitle,
+      icon: Database,
+      separator: true,
+    },
```
danad02, line 28:
```suggestion
    {
      title: cardTitle,
      icon: Database,
      separator: true,
    },
```

packages/ui/src/Card.tsx
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * ### Purpose
+ * Card components define content hierarchy (visual and logical), separating content into "groups".
+ *
+ * ### General Info
+ * - Card can be rendered with an optional separator and/or icon.
+ */
+export const Card: FC<CardProps> = ({ title, icon, separator = false, 'data-test': dataTest, children }) => (
+  <div data-test={dataTest ?? 'card-wrapper'} className={styles.wrapper}>
+    {title && (
```
aigoncharov, line 22: Seems like `title` is required. DO we need this `&&`?

packages/ui/src/Card.tsx
aigoncharov: nit: I feel like we could not add new elements to the DOM and simply use `border-bottom` on the `h3`. WDYT?

danad02: > Well, I think we can, but will it improve something? It's just a habit of mine to use span elements as separators.

It won't improve anyting visually or semantically. It's just that changing span's display to block is an extra step during implementation - div is a block element by default.

packages/ui/src/Card.module.scss
jozefhruska: Updated 👍

packages/ui/src/Card.tsx
jozefhruska: Updated, cc @aigoncharov 👍

packages/ui/src/Card.tsx
jozefhruska: Updated, thanks 🙏

packages/ui/__tests__/Card.test.tsx
jozefhruska: I just wanted to be consistent, but maybe readability is better, you're right :) I replaced all "0 length" checks with `exists()`, thanks!

packages/ui/src/Card.tsx
jozefhruska: @danad02 Replaced with `<div>` 👍

jozefhruska:
Thanks for great comments @danad02 and @aigoncharov 🙏  Resolved all of them, please re-review once you have a minute 😌

packages/ui/__tests__/Card.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const icon = wrapper.find(Icon)
+    const title = wrapper.findDataTest('card-heading')
+    const separator = wrapper.findDataTest('card-separator')
+
+    if (props.title) {
+      expect(title.text()).toEqual(props.title)
+      expect(title.getDOMNode().tagName).toEqual('H3')
+    } else {
+      expect(title.exists()).toBeFalsy()
+    }
```
aigoncharov, line 43:
```suggestion
      expect(title.text()).toEqual(props.title)
      expect(title.getDOMNode().tagName).toEqual('H3')
```

aigoncharov (approved):
LGTM once https://github.com/hazelcast/frontend-shared/pull/119/files?short_path=c8bbea7#r540820420 is fixed! GJ!

danad02 (approved):
LGTM. Nice job! :+1:

PR #120, opened 2020-12-10
Feat/toggle

utkukaratas:


packages/ui/__stories__/Toggle.stories.tsx
```diff
@@ -0,0 +1,214 @@
...
+export const WithHelperText = () => (
+  <Toggle helperText="this is an helper text." checked label="Label" name="default" onChange={(e) => logger.log('change', e.target.checked)} />
+)
+
+export const WithError = () => (
+  <Toggle error="this is an error message!" checked label="Label" name="default" onChange={(e) => logger.log('change', e.target.checked)} />
+)
+
+
+// export const FocusedUnchecked = () => (
```
aigoncharov, line 66: @utkukaratas is it a WIP PR?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,75 @@
...
+        onBlur={onBlur}
+        value={value}
+        disabled={disabled}
+        aria-checked={ariaChecked}
+        aria-invalid={!!error}
+        aria-describedby={helperText && helpTooltipId(id)}
+        aria-errormessage={error && errorId(id)}
+      />
+
+      <label className={cn(classNameLabel, {[styles.disabled]: disabled})} htmlFor={id}>
```
aigoncharov, line 64: Couldn't we use our Label?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,75 @@
...
+  helperText,
+  disabled,
+  className,
+  classNameLabel,
+  'data-test': dataTest,
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null)
+  const id = useUID()
+
+  let ariaChecked: AriaAttributes['aria-checked'] = checked ? 'true' : 'false'
```
aigoncharov, line 44: nit: why do we need to extract this variable? Couldn't we simply pass 'checked' to the aria-checked prop of the input?

aigoncharov: On second thought, why do we need aria-checked at all? We already have checked prop, shouldn't it be enough?

utkukaratas: hmm, i based in on Checkbox without giving it much thought. will check it out...

packages/ui/__stories__/Toggle.stories.tsx
utkukaratas: yep, this is a draft pr.

packages/ui/src/Toggle.tsx
aigoncharov: For checkbox it makes sense because checkbox could be `indeterminate`. It is not the issue with the toggle

packages/ui/src/Toggle.tsx
utkukaratas: unfortunately :/ i'm abusing the internal label element to display a track there.

packages/ui/src/Toggle.tsx
utkukaratas: yep, on a second look it's unnecessary. less LOC is best LOC. pushed the code. (ps. it's awesome to work with this level of peer reviews, thx a lot @aigoncharov, keep em coming!!)

packages/ui/src/Toggle.tsx
aigoncharov: `Label` accepts `ReactElement` as `label`. Couldn;t you pass anything you want, including your toggle track?

utkukaratas:
will reopen this.

PR #122, opened 2020-12-11
Wrap all radio stories in form

### LINKED ISSUE #121: Broken IDs in storybook (Radio input)
I'm not sure yet whether this is something wrong with Radio component or our new ID system, but after https://github.com/hazelcast/frontend-shared/commit/7104fd703599d8f564fc355d759f453c6501753f we have a problem with Storybook, visible in a Radio story (I'm not sure how this was not caught by visual checks)

Somehow IDs are messed up, because DOM acts in a very weird way. I've tried to reveal the native radio button and  checked that it's `checked`, but the visualization is wrong.

Screenshot 2020-12-11 at 9 48 51
Screenshot 2020-12-11 at 9 48 40

Since also native DOM behaves in a weird way (radio is checked, but not visualized that way), I assume it could have sth to do with invalid IDs in storybook. According to the spec, they should contain at least one letter https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#the-id-attribute. But `useUid` returns just plain numbers.

aigoncharov:
Closes #121

### DIFF
--- a/packages/ui/__stories__/Checkbox.stories.tsx
+++ b/packages/ui/__stories__/Checkbox.stories.tsx
@@ -2,13 +2,15 @@ import React, { useState } from 'react'
 import { Form, Formik } from 'formik'
 import { logger } from '@hazelcast/services'
 
-import styles from '../src/Checkbox.module.scss'
-
 import { Checkbox, CheckboxFormik, Link } from '../src'
+import { formDecorator } from './decorators'
+
+import styles from '../src/Checkbox.module.scss'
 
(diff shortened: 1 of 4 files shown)


jvorcak (approved):

jvorcak:
@aigoncharov it currently doesn't cause any issue, but maybe we could also wrap checkbox stories in order to prevent some future misunderstanding in case somebody adds more?

danad02 (approved):

Issue #118, opened 2020-12-10
Allow Button and Link component to change underlying element


jvorcak:
We need to be able to visually render the Button component, but allow `<a>` semantics.
The same with Link, something it's visualized as Link, but semantically it's a button.

Example:

<img width="337" alt="Screenshot 2020-12-09 at 19 07 16" src="https://user-images.githubusercontent.com/1083817/101760462-2cdf1b80-3adb-11eb-9209-471f642fcf98.png">

We need to investigate the code in order to come up with a fully type-safe solution to do this.

@aigoncharov 's comment from Slack
```I suggest we add an href property to the button. When it is passed, we render a instead of button. For the Link, we could make href optional. If href is undefined, we render a button instead of a. Would it cover the needs of the cloud?```

The same problem solved in MaterialUI: https://material-ui.com/api/button/

PR #123, opened 2020-12-11
Button/Link with a/button semantics

jvorcak:


packages/ui/src/Button.tsx
```diff
@@ -45,28 +45,47 @@ export type ButtonAccessibleIconRightProps =
...
+          component: 'button'
+          href?: never
+        }
+      | {
+          component?: undefined
+          href?: never
+        }
+    ) &
+      (ButtonDisabledProps | ButtonNotDisabledProps) &
+      Pick<ButtonHTMLAttributes<HTMLButtonElement>, 'autoFocus' | 'type'>)
```
danad02, line 86: This can be simplified if I'm not mistaken:
```suggestion
type ButtonTypeProps =
  | ({
      component: 'a'
      href: string
    } & ButtonNotDisabledProps)
  | ({
      component?: 'button'
      href?: never
    } & (ButtonDisabledProps | ButtonNotDisabledProps) &
      Pick<ButtonHTMLAttributes<HTMLButtonElement>, 'autoFocus' | 'type'>)
```

packages/ui/src/Button.tsx
```diff
@@ -45,28 +45,47 @@ export type ButtonAccessibleIconRightProps =
...
 
-export type ButtonProps = ButtonCommonProps &
-  ButtonAccessibleIconLeftProps &
-  ButtonAccessibleIconRightProps &
-  ButtonDisabledProps &
-  Pick<ButtonHTMLAttributes<HTMLButtonElement>, 'onClick' | 'className' | 'autoFocus' | 'type'>
+export type ButtonProps = ButtonCommonProps & ButtonAccessibleIconLeftProps & ButtonAccessibleIconRightProps & ButtonTypeProps
```
danad02, line 88: So we're allowing icons when we use `<a>` - how does it look? Do we want to allow it?

packages/ui/src/Button.tsx
```diff
@@ -45,28 +45,47 @@ export type ButtonAccessibleIconRightProps =
...
+}
 
 // Common props for all button "kinds"
 type ButtonCommonProps = {
   kind?: ButtonKind
   children: string
   capitalize?: boolean
   bodyClassName?: string
-}
+} & Pick<HTMLAttributes<HTMLAnchorElement | HTMLButtonElement>, 'className' | 'onClick'>
```
danad02, line 68: Do we want to allow usage of `onClick` for `<a>`?

danad02: Ah it's probably the same reason as the one in `Link` component:

> // it's also required by next.js for <a> https://nextjs.org/docs/api-reference/next/link

packages/ui/src/Button.tsx
jvorcak: Nope we don't, will fix, nice catch, I'll fix.

packages/ui/src/Link.module.scss
```diff
@@ -1,5 +1,10 @@
 @use '../styles/constants' as c;
 
+button.buttonReset {
```
aigoncharov, line 3: nit: Should we just include it into `@mixin link`? I mean it should not hurt `a`, right?

packages/ui/src/Link.module.scss
```diff
@@ -1,5 +1,10 @@
...
+  border: none;
+  background: none;
+}
+
 @mixin link {
```
aigoncharov, line 8: I just saw we use `outline: none;` for `:focus`. Shoud it be `outline: transparent solid c.$outlineWidth` instead?

packages/ui/src/Button.module.scss
```diff
@@ -1,7 +1,12 @@
 @use '../styles/constants' as c;
 @use '../styles/helpers' as h;
 
+a.button {
+  text-decoration: none;
```
aigoncharov, line 5: What if we just add it to the `.button`? It should not hurt the HTML `button`, right?

aigoncharov:
@jvorcak great prototype! Let's add more stories and tests, and merge this one in!

packages/ui/src/Button.tsx
jvorcak: @danad02 my bad, we actually do allow this :) Well visually it's still a button and you can use an icon.

jvorcak:
Closes #118

packages/ui/src/Button.tsx
```diff
@@ -110,11 +130,14 @@ export const Button = forwardRef<HTMLElement, ButtonProps>(
...
       iconRightColor,
       iconRightClassName,
+      rel = 'noopener',
+      target,
       ...rest
     },
     ref,
   ) => {
     const tooltipId = useUID()
+    const relFinal = useDeepCompareMemo(() => (Array.isArray(rel) ? rel.join(' ') : rel), [rel])
```
aigoncharov, line 140: @jvorcak I am not sure if `useMemo` makes sense here. The result is a string. There is no such thing as a reference to a string in JavaScript, right? So we can easily re-compute this string on every re-render and pass it to `Component`. If string value is the same it won;t cause the `Component` to re-render.

aigoncharov (approved):
GJ! Should be good to merge once https://github.com/hazelcast/frontend-shared/pull/123/files?short_path=5d114bf#r543139928 is resolved!

packages/ui/src/Button.tsx
jvorcak: @aigoncharov I think you're right, I took this rom `Link` component, but to be honest, I haven't thought of this more in depth. I assume I should remove this from Link also?

aigoncharov: @jvorcak probably, I did not think it through back then while implementing it in the `Link` :) My bad.

jvorcak: But I'm thinking ..., since the parameter can also be an array, memoization actually might make sense here. I'll follow up on Slack.

packages/ui/src/Button.tsx
danad02: yes, you're right :smile:

PR #125, opened 2020-12-15
Adds className property for Toast and Badge

jvorcak:
We were missing the possibility to add `className` for these two.


aigoncharov (approved):
GJ!
BTW, I was mistaken when I asked you to put the `className` override as the last class. As it turns out, the order of CSS classes in HTML does not matter. My bad.

PR #126, opened 2020-12-15
Feat/toggle

### LINKED ISSUE #18: [UI core library] Toggle
https://www.figma.com/file/8mVm6LTbp2Z0RaWWjTZoft/%F0%9F%90%9DHIVE---Hazelcast-Design-System?node-id=10187%3A30

        1. name
        2. checked
        3. onChange
        4. onBlur
        5. label
        6. error
        7. touched
2. Create a formik v2 wrapper

utkukaratas:
Closes #18


utkukaratas:
@aigoncharov i tried using Label component as the inner label but it looked messier than the current code which is like 3 lines anyway. all your other comments from the deleted pr should be handled by this pr.

aigoncharov:
@utkukaratas please, do not forget to connect the PR with the issue on zenhub and link to it via comment as well.

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,98 @@
+@use '../styles/constants' as c;
+@use '../styles/helpers' as h;
+
+$h: 24px;
```
aigoncharov, line 4: @utkukaratas as you can see in other components, all custom sizes are multipliers of `c.$grid`. This way all of our components will scale if we ever decide to change the grid. Moreover, `c.$grid` is in `rem`. So our UI scales for visually impaired users who increase their base font size

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,98 @@
...
+$w: 60px;
+$knob-r: $w / 2 + 8px;
```
aigoncharov, line 6: nit: Should we use cameCase for consistency with other components?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,98 @@
...
+$track-bg-on: #1dbc22;
+$track-bg-off: c.$colorNeutralLight; // rgba(0,0,0,.32);
+$knob-bg-on: c.$colorNeutralWhite;
+$knob-bg-off: c.$colorNeutralLighter;
+$label-margin: 12px;
+$knob-h: 20px;
+$knob-w: 20px;
+$knob-offset: 2px;
+
+@keyframes fadein {
```
aigoncharov, line 17: Great job with the animation!

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,98 @@
...
+
+  // th "ON" text
+  input[type='checkbox']:checked + label .toggle-track:after {
+    content: 'ON';
+    position: absolute;
+    top: 2px;
+    right: 8px;
+    font-size: 0.9rem;
+    color: #fff;
+    animation: fadein 0.2s;
```
aigoncharov, line 87: Should we create `styles/constants/_animations.scss` and add `0.2s` as `$transitionTimeMedium` there?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,77 @@
...
+        type="checkbox"
+        ref={inputRef}
+        id={id}
+        name={name}
+        checked={!!checked}
+        onChange={onChange}
+        onBlur={onBlur}
+        value={value}
+        disabled={disabled}
+        aria-checked={checked}
```
aigoncharov, line 56: Not needed since we have `checked`
```suggestion
```

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,77 @@
...
+  error,
+  checked,
+  label,
+  helperText,
+  disabled,
+  className,
+  classNameLabel,
+  'data-test': dataTest,
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null)
```
aigoncharov, line 40: What is it for?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  // global disabled state
+  .disabled .toggle-track {
+    opacity: 0.3;
```
aigoncharov, line 92: Should we also add `cursor: not -allowed`?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,77 @@
...
+        disabled={disabled}
+        aria-checked={checked}
+        aria-invalid={!!error}
+        aria-describedby={helperText && helpTooltipId(id)}
+        aria-errormessage={error && errorId(id)}
+      />
+
+      {/* label controlling the input above with the `toggle-track` element */}
+      <label className={cn(classNameLabel, { [styles.disabled]: disabled })} htmlFor={id}>
+        <span className={cn(styles.labelText)}>{label}</span>
```
aigoncharov, line 64:
```suggestion
        <span className={styles.labelText}>{label}</span>
```

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <span className={cn(styles['toggle-track'])}></span>
+      </label>
+
+      {helperText && <Help parentId={id} helperText={helperText} />}
+
+      {error && (
+        <div>
+          <Error error={error} className={styles.errorContainer} inputId={id} />
+        </div>
+      )}
```
aigoncharov, line 74: `Error` should always be mnounted. Otherwise `aria-live` inside of it won't work
```suggestion
        <div>
          <Error error={error} className={styles.errorContainer} inputId={id} />
        </div>
```

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,77 @@
...
+  onBlur?: (e: FocusEvent<HTMLInputElement>) => void
+  onChange?: (e: ChangeEvent<HTMLInputElement>) => void
+  error?: string
+
+  label: string | React.ReactNode
+  helperText?: string
+  disabled?: boolean
+  className?: string
+  classNameLabel?: string
+}
```
aigoncharov, line 22: Core props are the ones that are passed by `formik`. Extra props are everything else we need. So I would suggest to split it like this.
```suggestion
type ToggleCoreProps = {
  name: string
  value?: string
  checked?: boolean
  onBlur?: (e: FocusEvent<HTMLInputElement>) => void
  onChange?: (e: ChangeEvent<HTMLInputElement>) => void
  error?: string
}
export type ToggleExtraProps = {
  label: string | React.ReactNode
  helperText?: string
  disabled?: boolean
  className?: string
  classNameLabel?: string
}
type ToggleProps = ToggleCoreProps & ToggleExtraProps
```
Now We can use only `ToggleExtraProp` in `ToggleFormik`

packages/ui/src/ToggleFormik.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactElement } from 'react'
+import { useField } from 'formik'
+import { FieldValidatorGeneric, getFieldError } from './utils/formik'
+
+import { Toggle, ToggleProps } from './Toggle'
+
+export type ToggleFormikProps = ToggleProps & {
+  name: string
+  validate?: FieldValidatorGeneric<boolean>
+}
```
aigoncharov, line 10: Depends on https://github.com/hazelcast/frontend-shared/pull/126/files?short_path=b082a23#r543642690
```suggestion
export type ToggleFormikProps<V extends object> = ToggleExtraProps & {
  name: ExtractKeysOfValueType<V, boolean>
  validate?: FieldValidatorGeneric<boolean>
}
```
`ExtractKeysOfValueType` allows us to check that value of the prop passed as `name` is actually `boolean` in `formik`

aigoncharov (changes requested):
@utkukaratas GJ! Left a couple of comments

packages/ui/src/Toggle.module.scss
utkukaratas: done. (wish it was kebab case all around but too late :/)

packages/ui/src/Toggle.module.scss
utkukaratas: done.

packages/ui/src/Toggle.tsx
utkukaratas: done!

packages/ui/src/Toggle.tsx
utkukaratas: copy paste from checkbox :)) done.

packages/ui/src/Toggle.module.scss
utkukaratas: done.

packages/ui/src/Toggle.tsx
utkukaratas: done.

packages/ui/src/Toggle.tsx
utkukaratas: done.

packages/ui/src/Toggle.tsx
utkukaratas: done.

packages/ui/src/ToggleFormik.tsx
utkukaratas: done. (the type magic here is way over my head for now.)

utkukaratas:
thanks again for the great points @aigoncharov. all implemented except the sizing stuff - for that i'm waiting the final design.

aigoncharov:
> thanks again for the great points @aigoncharov. all implemented except the sizing stuff - for that i'm waiting the final design.

@utkukaratas AFAICS, @pawelklasa provided the design for the `off` state of the toggle and for the focused state. Let's follow the common sense here and replicate the design based on the `Checkbox` component:
- `hover` - use the same border style as specified for the Toggle  `focus` design, but use the color from checkbox hover.
- `disabled` - use all colors from the checkbox design
- `error` - same as hover, but use the error color
- `helper text` - use the same `Help` component with the same margin as for the checkbox.

packages/ui/__stories__/Toggle.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const WithError = () => (
+  <Toggle error="this is an error message!" checked label="Label" name="default" onChange={(e) => logger.log('change', e.target.checked)} />
+)
+
+export const ToggleWrappedInFormik = () => {
+  const validateTurbo = (value: boolean) => (!value ? 'turbo mode must be on!!!' : undefined)
+
+  const TestForm = () => (
+    <Formik
+      initialValues={{
```
aigoncharov, line 74: Please, provide a generic for the form value. See `CheckboxWrappedInFormik` story for example.

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+      {/* label controlling the input above with the `toggle-track` element */}
+      <label className={cn(classNameLabel, { [styles.disabled]: disabled })} htmlFor={id}>
+        <span className={styles.labelText}>{label}</span>
+        <span className={cn(styles['toggle-track'])}></span>
+      </label>
+
+      {helperText && <Help parentId={id} helperText={helperText} />}
+
+      <div>
```
aigoncharov, line 69: `Error` is a block itself. Why do we need the wrapping `div`?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,74 @@
...
+        disabled={disabled}
+        aria-invalid={!!error}
+        aria-describedby={helperText && helpTooltipId(id)}
+        aria-errormessage={error && errorId(id)}
+      />
+
+      {/* label controlling the input above with the `toggle-track` element */}
+      <label className={cn(classNameLabel, { [styles.disabled]: disabled })} htmlFor={id}>
+        <span className={styles.labelText}>{label}</span>
+        <span className={cn(styles['toggle-track'])}></span>
```
aigoncharov, line 64: Could you use cameCase for className for consistency? The accessing properties is much simpler
```suggestion
        <span className={styles.toggleTrack}></span>
```

utkukaratas:
hi @aigoncharov, all of your suggestions are taken into account with the latest commits.  (i did imitate Checkbox as much as it makes sense & all around refactoring to the css)

aigoncharov:
@utkukaratas GJ!
1. Could you make sure it passed the CI checks?
2. When the component is in an error state, the focus styles for the border should override the error ones.
image

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+$w: c.$grid * 15;
+$labelMargin: c.$grid * 4;
+$trackBr: 22px;
+$trackBgOn: #1dbc22;
+$trackBgOff: c.$colorNeutralLight;
+$knobRight: $w * 0.6;
+$knobBgOn: c.$colorNeutralWhite;
+$knobBgOff: c.$colorNeutralLighter;
+$knobR: $h * 0.9;
+$knobOffset: c.$grid * 0.3;
```
aigoncharov, line 14: In figma it is 2px. Our `$grid` is `0.25rem` which is an equivalent of `4px`. Should it be `0.5 * $grid`?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+    right: $knobRight;
+  }
+
+  // the "OFF" text
+  input[type='checkbox'] + label .toggleTrack:after {
+    content: 'OFF';
+    position: absolute;
+    top: 1px;
+    left: 6px;
+    font-size: $knobFontSize;
```
aigoncharov, line 123: We have typography mixins to handle fonts. Should we use `@typographyBodySmall` instead of the direct `font-size` assignment?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+    color: #fff;
```
aigoncharov, line 124: Should we replace it with `$colorNeutralWhite`?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+$h: c.$grid * 6;
+$w: c.$grid * 15;
+$labelMargin: c.$grid * 4;
+$trackBr: 22px;
+$trackBgOn: #1dbc22;
+$trackBgOff: c.$colorNeutralLight;
+$knobRight: $w * 0.6;
+$knobBgOn: c.$colorNeutralWhite;
+$knobBgOff: c.$colorNeutralLighter;
+$knobR: $h * 0.9;
```
aigoncharov, line 13: In figma, the knob radius is `20px`. It is `5 * $grid`. `$h * 0.9` is `5.4 * $grid`. Should we use `5 * $grid`?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
+@use '../styles/constants' as c;
+@use '../styles/helpers' as h;
+
+$h: c.$grid * 6;
+$w: c.$grid * 15;
+$labelMargin: c.$grid * 4;
+$trackBr: 22px;
+$trackBgOn: #1dbc22;
+$trackBgOff: c.$colorNeutralLight;
+$knobRight: $w * 0.6;
```
aigoncharov, line 10: Do we need this one at all? Could we `$knobOffset` as `right` or `left` instead depending on the state?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
+@use '../styles/constants' as c;
+@use '../styles/helpers' as h;
+
+$h: c.$grid * 6;
+$w: c.$grid * 15;
+$labelMargin: c.$grid * 4;
+$trackBr: 22px;
+$trackBgOn: #1dbc22;
```
aigoncharov, line 8: Let's use `$colorSuccessLight` for consistency with our color palette. We'll ask @pawelklasa to verify once we implment it.

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
+@use '../styles/constants' as c;
+@use '../styles/helpers' as h;
+
+$h: c.$grid * 6;
+$w: c.$grid * 15;
+$labelMargin: c.$grid * 4;
```
aigoncharov, line 6: In figma, `margin-right` is `10px`. Should it be `$grid * 2.5` then?

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+$trackBr: 22px;
```
aigoncharov, line 7:
```suggestion
$trackBr: 50%;
```

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+}
+
+@mixin focusState {
+  @include h.outlineFormField;
+}
+
+.wrapper {
+  // hide the actual input, we will show the toggleTrack instead.
+  input[type='checkbox'] {
+    display: none;
```
aigoncharov, line 37: I would suggest using `@visuallyHidden` instead of `display: none` to keep it accessible for the screen readers

packages/ui/src/Toggle.module.scss
```diff
@@ -0,0 +1,138 @@
...
+    background: $knobBgOn;
+    right: $knobRight;
+  }
+
+  // the "OFF" text
+  input[type='checkbox'] + label .toggleTrack:after {
+    content: 'OFF';
+    position: absolute;
+    top: 1px;
+    left: 6px;
```
aigoncharov, line 122: AFAICS, the idea is to keep text visually centered. Could we do this to avoid calculating pixels?
```suggestion
    top: 50%;
    left: calc(50% - #{${knobR}});
    transform: translate(-50%, -50%);
```

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+          },
+          classNameLabel,
+        )}
+        htmlFor={id}
+      >
+        <span className={styles.labelText}>{label}</span>
+        <span className={styles.toggleTrack} aria-label={name} aria-checked={!!checked} role="checkbox" tabIndex={0}></span>
+      </label>
+
+      {helperText && <Help parentId={id} helperText={helperText} />}
```
aigoncharov, line 76: In figma, for the checkbox and for the radio, the helper icon has a right offset of around `10px`. The helper icon has padding of $grid. Should we add margin-left of `1.5 * $grid`?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+        id={id}
+        name={name}
+        checked={!!checked}
+        onChange={onChange}
+        onBlur={onBlur}
+        value={value}
+        disabled={disabled}
+        aria-invalid={!!error}
+        aria-describedby={helperText && helpTooltipId(id)}
+        aria-errormessage={error && errorId(id)}
```
aigoncharov, line 58: Should we move all of these to the span with `styles.toggleTrack` since it is acting as our checkbox?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+          {
+            [styles.error]: !!error,
+            [styles.disabled]: disabled,
+          },
+          classNameLabel,
+        )}
+        htmlFor={id}
+      >
+        <span className={styles.labelText}>{label}</span>
+        <span className={styles.toggleTrack} aria-label={name} aria-checked={!!checked} role="checkbox" tabIndex={0}></span>
```
aigoncharov, line 73: Why `aria-label` is `name`? Should it be `aria-label={label}`?

aigoncharov, line 73: Should we add `aria-disabled` here since we are using it instead of the checkbox?

packages/ui/src/Toggle.tsx
```diff
@@ -0,0 +1,81 @@
...
+      {/* hidden yet actual input */}
+      <input
+        type="checkbox"
+        id={id}
+        name={name}
+        checked={!!checked}
+        onChange={onChange}
+        onBlur={onBlur}
+        value={value}
+        disabled={disabled}
```
aigoncharov, line 55: Disabled inputs should not receive focus for consistency with our other inputs. Try focusing the disabled radio or checkbox.

packages/ui/src/Toggle.module.scss
utkukaratas: nope, that just makes it too round. setting to resolved.

packages/ui/src/Toggle.tsx
utkukaratas: label can be a ReactNode so we cannot pass it there.

utkukaratas: good idea, indeed i will move every aria attribute to the span element.

aigoncharov: Should we add an `id` to our label then and use `aria-labelledby`?

packages/ui/src/Toggle.tsx
utkukaratas: checkbox does not give an explicit margin and looks the same as toggle. if we were to be pixel perfect with the design we should do both then?

packages/ui/src/Toggle.module.scss
utkukaratas: yup cleaner. implemented.

packages/ui/src/Toggle.module.scss
utkukaratas: that seems too interdependent. this way is better imo.

packages/ui/src/Toggle.module.scss
utkukaratas: implemented the vertical one, for horizontal we now use $grid.

packages/ui/src/Toggle.module.scss
utkukaratas: nope, .5 looks way off. .3 is the magic number.

packages/ui/src/Toggle.module.scss
utkukaratas: great catch. fixed without magic constants.

packages/ui/src/Toggle.tsx
utkukaratas: aria-labelledby is done.

utkukaratas:
@aigoncharov thanks for another great round of reviews, extremely helpful for me. 

i implemented most of your suggestions and resolved related comments. the comments i left unresolved are the ones i had other ideas. feel free to resolve them or pressure me to look again.

ps. i'm still pursuing the pr errors with clean install...

packages/ui/src/Toggle.module.scss
aigoncharov: ` @typographyBodySmall` sets not only the font-size, but also line-height, font-family, font-weight and letter-sapcing. It is a uniform way to work with typography in the core UI. What downsides do you see?

packages/ui/src/Toggle.tsx
aigoncharov: Yeah. We have bug in `Checkbox` component. I'll raise a PR for it later.

aigoncharov:
@utkukaratas I have merged `master` into your branch, ran `npx lerna clean`, `npm ci`, and it started giving correct errors.
Here they are:
```
andreygoncharov@andrey-goncharov-ThinkPad-X1-Extreme:~/Dev/Hazelcast/frontend-shared/packages/ui$ npm run lint

> @hazelcast/ui@0.1.0-alpha.0 lint /home/andreygoncharov/Dev/Hazelcast/frontend-shared/packages/ui
> npm run lint:ts && npm run lint:css


> @hazelcast/ui@0.1.0-alpha.0 lint:ts /home/andreygoncharov/Dev/Hazelcast/frontend-shared/packages/ui
> eslint --ext .ts,.tsx .


> @hazelcast/ui@0.1.0-alpha.0 lint:css /home/andreygoncharov/Dev/Hazelcast/frontend-shared/packages/ui
> postcss --config ./.postcss-lint --replace '**/*.{css,scss}'


src/Toggle.module.scss
92:3	⚠  Expected selector ".wrapper .error .toggleTrack" to come before selector ".wrapper .disabled .toggleTrack:focus" (no-descending-specificity) [stylelint]
104:15	⚠  Expected double colon pseudo-element notation (selector-pseudo-element-colon-notation) [stylelint]
118:54	⚠  Expected double colon pseudo-element notation (selector-pseudo-element-colon-notation) [stylelint]
124:46	⚠  Expected double colon pseudo-element notation (selector-pseudo-element-colon-notation) [stylelint]
136:54	⚠  Expected double colon pseudo-element notation (selector-pseudo-element-colon-notation) [stylelint]
139:5	⚠  CSS unset value not supported by: IE (11) (css-unset-value) [doiuse]

Error: 
** postcss-reporter: warnings or errors were found **

```

aigoncharov:
@utkukaratas did you have a chance to take a look at the border color thing for the focused toggle in the error state?

packages/ui/src/Toggle.module.scss
aigoncharov: I see. Because of the border that takes `1px`, the offset is `1px` not `2px`. Should it be `c.$grid * 0.25` then? Otherwise. it is calculated as `1.2px`.

utkukaratas:
@aigoncharov a clean install did the trick, weird. anyhow this branch should now pass from the pr builder.

> @utkukaratas did you have a chance to take a look at the border color thing for the focused toggle in the error state?

yes this should be fixed.

packages/ui/src/Toggle.module.scss
utkukaratas: just a semantic issue IMO -- the on/off label here is nothing to do with general typography of the lib. it's just a static text glued there for looking nice.

packages/ui/__stories__/Toggle.stories.tsx
```diff
@@ -0,0 +1,133 @@
+import React, { useState } from 'react'
+import { Form, Formik } from 'formik'
+import { logger } from '@hazelcast/services'
+import styles from '../src/Toggle.module.scss'
+import { Toggle, ToggleFormik } from '../src'
+
+export default {
+  title: 'Components/Toggle',
+  component: Toggle,
```
aigoncharov, line 9:
```suggestion
  component: Toggle,
  decorators: [formDecorator],
```

packages/ui/src/Toggle.module.scss
aigoncharov: Shouldn't this text look consistent with any other text in any other component of the lib? That is the goal of the `@typography` helpers

aigoncharov (approved):

PR #127, opened 2020-12-17
Feat/table

danad02:
Linked to #9 

What is included in this PR:

- A `Pagination` component
    - Includes `usePagination` hook, stories and bunch of tests. 

- A `Table` component
    - Custom `Row`, `Cell`, `Header` and `Footer` components.
    - Includes column resizing, sorting, custom cell renderers and a possibility to add some custom cell props. 
    - Stories and tests included.

As this is a truly big PR I suggest you to go through it like this:
Start with the `Pagination` component and its `usePagination` hook. Take a look at stories and tests to better understand if anything's unclear. Or just ask a question. The component is not bind to the `Table`, it's standalone and can be used anywhere.

When you're done with the `Pagination` you can start with the `Table`. Our `Table` is using the [`react-table`](https://react-table.tanstack.com/) package. I suggest to take a look at least at [basic example](https://react-table.tanstack.com/docs/examples/basic) to understand what's going on. Then you can slowly go through our own implementation. `Row`, `Cell`, `Header` and `Footer` components are very simple, you might want to start there and then put it all together. Comments, stories and tests might help you understand if something's unclear. 

I imagine there will be a lot of questions and suggestions so it might take a while to merge this in.


packages/ui/@types/react-table.d.ts
```diff
@@ -0,0 +1,126 @@
+import type {
```
aigoncharov, line 1: Could you elaborate on why we need this override?

aigoncharov:
@danad02 I have found a small bug.
Steps to reproduce:
1. Strat shrinking a column 
2. Move your mouse past a point of a reasonable min width

Expected result:
COlumn stays at its min width

Current result:
Column expands back to its original dimensions

image

aigoncharov:
@danad02 could we do something about the outline? I would suggest to make it purple, like everywhere else, and make it cover an entire cell
image

aigoncharov:
@danad02 regarding sorting:
1. Should we hide a chevron when the column is not sorted? 
2. Do you think we could somehow style the table header cells for hover, focus and active states when they are clickable? Could you ping @pawelklasa to get the designs?

aigoncharov:
@danad02 the pagination widget looks stunning! Great job! Do you think we could dynamically adjust it to different table sizes? In its current form it does not work well with smaller tables
ezgif-3-912a9c180b61

I would suggest to do the following for smaller table sizes:
1. If the pagination widget does not fit, remove `previous` and `next` text from the button, leaving only arrows.
2. If it still does not fit, decrease the number of visible pages. Keep removing them until we have [`N`, `N+1`, `...`, `Last`]. If `Last` or `Last -1` is currently selected, make it [`N`, `...`, `Last -1`, `Last`]
3. If it still does not fit, hide `rows per page`, `go to`. Add `gear` icon. When `gear` icon is clicked replace the row with pagintion with another row with the removed controls: `rows per page`, `go to`. Display `X` icon. When `X` icon is clicked, hide the currently displayed row of extra controls, and bring back the page numbers with arrows.

WDYT?

If you like it, I say we implement it and verify with Pawel after it. It's better to ask for forgiveness than permission :) However,  feel free to verify the current strategy with @pawelklasa  first. Your choice.

packages/ui/package.json
```diff
@@ -49,6 +49,7 @@
     "@popperjs/core": "^2.5.3",
     "classnames": "^2.2.6",
     "formik": "^2.2.0",
+    "namor": "^2.0.2",
```
aigoncharov, line 52: SHould it be a dev dependency?

packages/ui/src/Table/Row.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return (
+      <div data-test="table-header-row" className={cn(styles.headerRow, className)} style={style} role={role} aria-rowindex={ariaRowIndex}>
+        {children}
+      </div>
+    )
+  }
+
+  return (
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events
+    <div
```
aigoncharov, line 30: Do you think we could try adding `rowHref` prop of `(rowInfo: RowType<D>) => string` and make `Row` an `a` if  `rowHref` is present? I, personally, like clicking on the entire row instead of trying to find a clickable item inside of it. WDYT?
We could allow devs specify either `onRowClick` or `rowHref`.

packages/ui/src/Table/Table.tsx
```diff
@@ -0,0 +1,255 @@
...
+   * An example can be found in `ClickableRowsWithNameLink` story in Table.stories.tsx
+   */
+  onRowClick?: (rowInfo: RowType<D>) => void
+  // Custom props getter for Cell
+  getCustomCellProps?: (cellInfo: CellType<D>) => CellProps
+} & DataTestProp
+
+type TableProps<D extends object> = TableOptions<D> & ExtendedPaginationProps & ControlledPaginationProps & CustomTableProps<D>
+
+const column = {
```
aigoncharov, line 61: Could we add it to `styles/constants/sizes` and import from there?

packages/ui/src/Table/Table.tsx
```diff
@@ -0,0 +1,255 @@
...
+  // When using the useFlexLayout:
+  minWidth: 30, // minWidth is only used as a limit for resizing
+  width: 150, // width is used for both the flex-basis and flex-grow
+  maxWidth: 200, // maxWidth is only used as a limit for resizing
+}
+
+// Inspiration here: https://react-table.tanstack.com/docs/examples/basic
+export function Table<D extends object>({
```
aigoncharov, line 69: nit: should it be an arrow function for consistency?
BTW, should we export it from `src/index.tsx`?

danad02: Yes, and yes :+1:

packages/ui/src/Table/Table.tsx
danad02: Of course :+1:

danad02:
@aigoncharov About the `Column expands back to its original dimensions` bug - https://github.com/hazelcast/frontend-shared/pull/127#issuecomment-752925873.
This is a bug in `react-table` and it seems to be a known one: https://github.com/tannerlinsley/react-table/issues/2830.
You can actually check the same behavior in this example: https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/full-width-resizable-table

packages/ui/@types/react-table.d.ts
danad02: I added the link exaplaining to the top: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-table.
In a nutshell: the types depend upon declaration merging to work well.

packages/ui/src/Table/Table.tsx
danad02: Changed here: https://github.com/hazelcast/frontend-shared/pull/127/commits/8dc56f76951dcc6380237547b05fb5cfc24fe457

packages/ui/src/Table/Table.tsx
danad02: Changed in https://github.com/hazelcast/frontend-shared/pull/127/commits/db5631df6a459ca3d5aee70276c7381eb45182ed

packages/ui/src/Table/Row.tsx
danad02: That's worth a shot. Let me try this approach!

packages/ui/package.json
danad02: Moved to dev dependencies.

packages/ui/src/Table/Table.tsx
danad02: Upon trying this approach I didn't change the current implementation. The `width` have to be numbers but imported values from scss are strings, so we would have to parse those values and I don't like that solution.

packages/ui/src/Table/Row.tsx
danad02: `Table` now has two mutually exclusive props: `onClick` and `getHref`
```
onRowClick?: (rowInfo: RowType<D>) => void
getHref?: (rowInfo: RowType<D>) => string
```

`getHref` has to return string which is then used as `href` in our `<a>` element.

packages/ui/src/Table/Table.tsx
aigoncharov: I would rather parse strings then split constants across multiple unrelated files

danad02: Done :+1:

danad02:
I had a little chat and short call with @pawelklasa about [pagination's width](https://github.com/hazelcast/frontend-shared/pull/127#issuecomment-752932835).
The conclusion is to either show full pagination on one line and if there's not enough space then do the following:

1. Reduce number of page number buttons to 3 (instead of default 5).
2. Put Rows per page select and Page jump input under a button. Ideally, we would show overlay with these features after clicking the button. Since we don't have any component like that I had to improvise - upon button click we show those features instead of pages. Once we implement and overlay we should replace this a-bit-clunky solution.

aigoncharov (approved):
Outstanding work!

PR #131, opened 2020-12-19
Add default button type

### LINKED ISSUE #130: Idea: Required type property for a button
Right now we have `type` property optional on a Button. Should we make it required in order to force programmers to think about a11y and usage?

It's also recommended https://www.w3schools.com/tags/att_button_type.asp

aigoncharov:
Closes #130

@jvorcak I would say that 90% of our buttons are of `button` type. So I would not require the `type` property, but rather have a sane default.WDYT?


jvorcak:
@aigoncharov 👍  I'm ok with it.

The most common bug is that somebody places a `button` inside a form and the default is `submit` for most browsers. So having this default as `button` is ok.

jvorcak (approved):

jvorcak:
I'll merge if you're ok with it, it'd help me to avoid writing some code for my current work. Thanks a lot!

PR #137, opened 2020-12-22
feat(ui/icon): add support for simple-icons

### LINKED ISSUE #135: Support for social icons
Our current icons package `feather-icons` doesn't have "Google" icon we need for social sign in on Cloud. 
https://github.com/feathericons/feather/issues/763#issue-505003395

We need to extend our APIs to support at least "Google" icon.

jozefhruska:
Closes #135


packages/ui/__stories__/Icon.stories.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
 import { Clipboard } from 'react-feather'
 
-import { Icon } from '../src/Icon'
+import { Icon } from '../src'
+import { Google } from '@icons-pack/react-simple-icons'
```
aigoncharov, line 5: nit: I would suggest putting absolute imports first

packages/ui/__stories__/Icon.stories.tsx
```diff
@@ -10,12 +11,24 @@ export default {
   component: Icon,
 }
 
-export const Normal = () => <Icon icon={Clipboard} ariaLabel="Icon copy to clipboard" />
+export const FeatherIconNormal = () => <Icon icon={Clipboard} ariaLabel="Icon copy to clipboard" />
```
aigoncharov, line 14: Could you remove the screenshots for the old stories?

aigoncharov (approved):
LGTM!

packages/ui/__stories__/Icon.stories.tsx
jozefhruska: Sure, I missed this! Thanks

packages/ui/__stories__/Icon.stories.tsx
jozefhruska: Sorry, thought it does that automatically.

aigoncharov: No worries. I think we should contribute that cleanup functionality to Loki at some point

PR #138, opened 2020-12-22
fix(ui/textfield): formik version type options

### LINKED ISSUE #133: [UI core library] TextFieldFormik doesn't accept types other than "text"

jozefhruska:
Closes #133


packages/ui/src/TextFieldFormik.tsx
```diff
@@ -1,11 +1,11 @@
 import React, { ReactElement } from 'react'
 import { useField } from 'formik'
 
-import { TextField, TextFieldExtraProps } from './TextField'
+import { TextField, TextFieldExtraProps, TextFieldTypes } from './TextField'
 import { FieldValidatorGeneric, getFieldError } from './utils/formik'
 import { ExtractKeysOfValueType } from './utils/types'
 
-export type TextFieldFormikProps<V extends object> = Exclude<TextFieldExtraProps<'text'>, 'type'> & {
+export type TextFieldFormikProps<V extends object> = TextFieldExtraProps<TextFieldTypes> & {
```
aigoncharov, line 8: We should not allow `number` as the type. I think we should allow `'email' | 'search' | 'tel' | 'text' | 'url' | undefined` only.

BTW, we should pass the proper type on L22

jozefhruska: @aigoncharov Wow, good catch! But why shouldn’t we allow `number` type?

aigoncharov: @jozefhruska we have `NumberField` for it, right? And for `password` we have `PasswordField`.

jozefhruska: Cool, now I understand 👌 Will make the change, thanks.

aigoncharov (approved):

PR #139, opened 2020-12-22
fix(ui/passwordfield): input icon

### LINKED ISSUE #134: [UI core library] PasswordField doesn't support "inputIcon" prop

jozefhruska:
Closes #134


packages/ui/src/PasswordField.tsx
```diff
@@ -87,6 +87,7 @@ export const PasswordField: FC<PasswordFieldProps> = ({
       type={visible ? 'text' : 'password'}
       inputContainerChild={overlay}
       inputClassName={cn(styles.inputContainer, inputClassName)}
+      inputIcon={Lock}
```
aigoncharov, line 90: We do not have it in MC. At least, currently. Should we pass the `inputIcon` as prop? Then we could keep all the stories as is and add new ones with the icon present.

jozefhruska: @aigoncharov Would it even make sense to use any other icon than "Lock" in here? @pawelklasa maybe we could unify this between the Cloud (lock icon on password field) and MC (no icon on password field)?

aigoncharov: @jozefhruska we cannot really add the Lock icon for some fields in MC as it would require adding icons for all icons of the form for consistency and it just does not make sense for complex security provider configs.
Agreed though, that it is not likely to be anything else bu the `Lock`. Should we use a boolean prop then? Say, `withIcon`?

jozefhruska: @aigoncharov Yep, `withIcon` prop sounds reasonable. I'll make the change 👌

aigoncharov (approved):

packages/ui/src/PasswordField.tsx
```diff
@@ -25,6 +25,7 @@ type PasswordFieldCoreProps = {
   value?: string
   onBlur?: (e: FocusEvent<HTMLInputElement>) => void
   onChange: (e: ChangeEvent<HTMLInputElement>) => void
+  withIcon?: boolean
```
aigoncharov, line 28: Should it be in `PasswordFieldExtraProps`? Otherwise, we won;t be able to pass it in from the `PasswordFieldFormik`

jozefhruska: Indeed, I've updated it. Thanks :)

PR #143, opened 2020-12-24
42 close with esc

### LINKED ISSUE #42: Close `Toast` and `Alert` with `ESC`

utkukaratas:
closes #42 

there're 2 ways to handle this issue:

1. the straightforward OO way -- introduce a `ModalManager` manager object to track down anything that pops up (like Toast, Alert or actual Modal dialogs) and then hook&process the Escape key events globally. I haven't gone this way as it would complicate things too much at this point. maybe we could go this way after 1.0 release where modals and related components are stable.

2. a neater way of doing things is introducing a small hook `useCloseByEscKey` for `Escape tracking components` to use. this hook does hook the keydown event globally and calls the given callback to do whatever the caller wants when Escape is hit.

IMO the second way is more react-y and less magic so here it is.


packages/ui/src/utils/hooks/useCloseByEscKey.ts
```diff
@@ -0,0 +1,42 @@
...
+const _cbStack: CallbackFunction[] = []
+
+function _removeFromStack(cb: CallbackFunction) {
+  const idx = _cbStack.indexOf(cb)
+  if (idx > -1) {
+    _cbStack.splice(idx, 1)
+  }
+}
+
+export default function useCloseByEscKey(cb: CallbackFunction) {
```
aigoncharov, line 15: @utkukaratas 
1. Good thinking with the callback stack, however, wouldn't `stopImmediatePropagation` do the job?
2. We tend to use arrow functions everywhere. Should we do it here as well for consistency?
3. We tend to use named exports instead of default ones. Should we do it here for consistency?
4. We already depend on `react-use`. Should we consider using `useKey` from there?

utkukaratas: (2) and (3) (4)👍 

(1) `.stopImmediatePropagation` will swallow the event and that might meddle with some other components' event handling. not a conservative choice.

so i'm pushing a version that uses `useKey` underneath. if you think `.stopImmediatePropagation` would be okay then i will get rid of useCloseByEsc module all together and use `useKey` directly in Alert and Toast. up to you guys (cc @aigoncharov @danad02).

packages/ui/src/utils/hooks/useCloseByEscKey.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+}
+
+export const useCloseByEscKey = (cb: CallbackFunction) => {
+  // 'Esc' -> IE/Edge specific value
+  const predicate = (event: KeyboardEvent) => event.key === 'Escape' || event.key === 'Esc'
+  useKey(predicate, (nativeEvent: KeyboardEvent) => {
+    // check if the last callback on the stack is this cb.
+    const lastCb = _cbStack.length ? _cbStack[_cbStack.length - 1] : null
+    if (cb && lastCb === cb) {
```
aigoncharov, line 21: @utkukaratas What if `cb` changes?
Consider this case:
```tsx
const Component = () => {
  useCloseByEscKey(() => { /* do something */ })

  return ...
}
```
With every render of the `Component` we will create a new callback that will never be removed.

I suggest we create a `cbRef` with `useRef` and use `cbRef.current` inside of our own callback that we execute on kye press.
We should also add `useEffect` and remove old `cb` from the stack if it ever changes.
WDYT?

packages/ui/src/utils/hooks/useCloseByEscKey.ts
```diff
@@ -0,0 +1,32 @@
...
+      // if so, invoke it and remove it from the stack.
+      _removeFromStack(cb)
```
aigoncharov, line 23: nit: Could we simply `_cbStack.pop()`?

packages/ui/src/utils/hooks/useCloseByEscKey.ts
```diff
@@ -0,0 +1,32 @@
...
+function _removeFromStack(cb: CallbackFunction) {
+  const idx = _cbStack.indexOf(cb)
+  if (idx > -1) {
+    _cbStack.splice(idx, 1)
+  }
+}
+
+export const useCloseByEscKey = (cb: CallbackFunction) => {
+  // 'Esc' -> IE/Edge specific value
+  const predicate = (event: KeyboardEvent) => event.key === 'Escape' || event.key === 'Esc'
```
aigoncharov, line 17: Should we move `predicate` out of the render function?
`useKey` uses it as a dependency for the internal `useMemo` - https://github.com/streamich/react-use/blob/master/src/useKey.ts#L24

packages/ui/src/utils/hooks/useCloseByEscKey.ts
```diff
@@ -0,0 +1,32 @@
...
+const _cbStack: CallbackFunction[] = []
+
+function _removeFromStack(cb: CallbackFunction) {
+  const idx = _cbStack.indexOf(cb)
+  if (idx > -1) {
+    _cbStack.splice(idx, 1)
+  }
+}
+
+export const useCloseByEscKey = (cb: CallbackFunction) => {
```
aigoncharov, line 15: Should we move it to `packages/hooks`? I think we could re-use it in our app

packages/ui/src/utils/hooks/useCloseByEscKey.ts
aigoncharov: @utkukaratas solid argument about `stopImmediatePropagation`! Let's try to go with the stack like you suggested.

danad02: I kind of started thinking about some corner cases. Say you have a dialog (modal) opened. Usually you are able to close it by hitting esc key. What happens if we have a dialog opened, an alert is present and we hit the esc key? Should we even care about these cases?

aigoncharov: @danad02 good thinking! 
@utkukaratas I suggest we stick with `stopImmediatePropagation` after all. It is simpler and might do more good than bad.

danad02: We can add an option to **not** stop event propagation if we find it's needed or useful in some kind of scenario.

utkukaratas:
i've implemented all of the above suggesions guys, got rid of the useCloseByEscKey hook and inlined it into the components themselves. do please review from scratch as this amounts a completely new implementation. @aigoncharov @danad02 

ps. how about a `.dismissableByEscKey` prop to `Alert` and `Toast`? @danad02

danad02:
@utkukaratas The prop `dismissableByEscKey` sounds good, I like it :+1:

packages/ui/__tests__/Toast.test.tsx
```diff
@@ -67,4 +67,56 @@ describe('Toast', () => {
 
     expect(closeToast).toBeCalledTimes(1)
   })
+
```
aigoncharov, line 70: Good job with tests!

packages/ui/src/Alert.tsx
```diff
@@ -41,12 +43,21 @@ export type AlertProps = {
...
 
+  useKey(
+    escKeyFilterPredicate,
+    (nativeEvent: KeyboardEvent) => {
+      closeToast && closeToast()
+      nativeEvent.stopImmediatePropagation()
+    },
+    { options: { once: true } },
+  )
+
```
aigoncharov, line 60: `useKey` uses [predicate] as  deps by default. What if `closeToast` changes at some point?

```suggestion
  const closeByEsc = useMemo(
    (nativeEvent: KeyboardEvent) => {
      closeToast?.()
      nativeEvent.stopImmediatePropagation()
    }, [closeToast]
  )
  
  useKey(
    escKeyFilterPredicate,
    closeByEsc,
    { options: { once: true } },
    [closeByEsc]
  )

```

Same goes for the Toast

utkukaratas: done.

packages/ui/src/Alert.tsx
```diff
@@ -43,21 +43,23 @@ export type AlertProps = {
...
+export const Alert: FC<AlertProps> = ({ type, title, content, actions, className, dismissableByEscKey = true, closeToast }) => {
   const { icon, ariaLabel } = ToastIcon[type]
 
-  useKey(
-    escKeyFilterPredicate,
+  const closeByEsc = useCallback(
     (nativeEvent: KeyboardEvent) => {
-      closeToast && closeToast()
+      dismissableByEscKey && closeToast?.()
       nativeEvent.stopImmediatePropagation()
```
aigoncharov, line 56: @utkukaratas should we stopImmediatePropagation if the alert is not dismissable by esc?

utkukaratas: aaaaaaah! you've one hell of an eye for code andrey!!! fix incoming...

aigoncharov (approved):
LGTM!

PR #144, opened 2020-12-28
Feature/react select autocomplete off

jvorcak:
I admit this is not very elegant, but it works. Instead of using currently hard-coded value off, we hard-code it to chrome-off  which really disables it in all browsers.

Any other suggestions are welcome.

Do you have any other use-cases where we would actually need to provide a value for this component?

Problem with a current implementation:
Screenshot 2020-12-28 at 12 00 36

Related issue: https://github.com/JedWatson/react-select/issues/3500


packages/ui/src/SelectField.tsx
```diff
@@ -24,6 +24,13 @@ const ClearIndicator = ({ innerProps }: IndicatorProps<SelectFieldOption<any>>)
   return <IconButton {...innerProps} icon={X} ariaHidden kind="primary" size="normal" className={styles.clear} tabIndex={-1} />
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const Input = (props: any) => {
+  // autoComplete='off' is hard-coded inside SelectField, but doesn't work in Chrome.
+  // Having an invalid value hard-coded disabled it in all browsers.
+  return <components.Input {...props} autoComplete="one-time-code" />
```
aigoncharov, line 31: Why not `chrome-off`?

jvorcak: @aigoncharov I couldn't decide whether it's better to use an invalid a11y value or a value that shouldn't in practice effect functionality. I was 50/50, now I'm thinking an explicit chrome-off could be better, even though it's invalid.

aigoncharov (approved):

danad02 (approved):

jvorcak:
If it's ok with you I'll merge and fix https://github.com/hazelcast/frontend-shared/issues/145 afterward, I'm quite limited on time right now.

Issue #132, opened 2020-12-19
Double tooltip shown for a button


jvorcak:
![Screenshot 2020-12-19 at 15 44 20](https://user-images.githubusercontent.com/1083817/102692076-4c7bef80-4211-11eb-9e5a-008d15d5316a.png)

Hi,

I've discovered this during development, before I managed to reserve more space for the button.

Consider a button that has truncated text and is disabled in the same time. We should try to avoid showing 2 tooltips at once.

aigoncharov:
@jvorcak I suggest we should only the disabled tooltip. WDYT?

PR #147, opened 2021-01-07
Fix/double button tooltip

jvorcak:
Fixes https://github.com/hazelcast/frontend-shared/issues/132


jankoritak (approved):

packages/ui/src/Button.tsx
```diff
@@ -178,7 +178,16 @@ export const Button = forwardRef<HTMLElement, ButtonProps>(
                   color={iconLeftColor}
```
danad02, line 178: I'm wondering why we need both the `disabledTooltip` and the `disabledTooltipVisible` props. When we don't provide `disabledTooltip` prop then the `Tooltip` just doesn't render anything. @jankoritak @jvorcak Do you remember why we have to provide `disabledTooltipVisible` explicitly like this? Is this maybe an optimization?
If we just had the `disabledTooltip` prop it would simplify the flow.

packages/ui/src/Button.tsx
```diff
@@ -178,7 +178,16 @@ export const Button = forwardRef<HTMLElement, ButtonProps>(
...
-              <TruncatedText text={capitalize ? children.toUpperCase() : children} />
+              <TruncatedText
+                text={capitalize ? children.toUpperCase() : children}
+                /*
+                  If a button is disabled and text is long we don't want to show 2 popups.
+                  1. In case a button is disabled and no `disabledTooltipVisible` is enforced, we just hide truncated popup.
+                  2. In case a button is disabled and disabledTooltip is hidden with a false flag, there is a space to show
+                  the truncated popup -> setting it to undefined will leave the default behaviour.
+                */
+                tooltipVisible={disabled && disabledTooltipVisible !== false ? false : undefined}
```
danad02, line 189: nit: a bit more readable
```suggestion
                tooltipVisible={disabled && disabledTooltipVisible ? false : undefined}
```

packages/ui/__tests__/Button.test.tsx
```diff
@@ -117,6 +118,40 @@ describe('Button', () => {
...
+        </Button>
+      </div>,
+    )
+
+    expect(wrapper.find(TruncatedText).props()).toMatchObject({
+      tooltipVisible: false,
+    })
+  })
+
+  it('Renders disabled button with a disabled tooltip and disabledTooltip is invisible, correct tooltipVisible flag passed to TruncatedText', async () => {
```
danad02, line 138: nit: This description is a bit hard to read :smile: 
How about something simple?
`Renders disabled button with invisible tooltip`

danad02 (approved):

packages/ui/src/Button.tsx
jvorcak: @danad02 this needs to stay the old way. It's important for us to check for an explicit `false` value.  What is tricky here is that `disabledTooltipVisible` can be `undefined`, there is a test for it.

jvorcak: What is tricky here is that `disabledTooltipVisible` can be `undefined`. There is a test for it.

packages/ui/__tests__/Button.test.tsx
jvorcak: Adjusted

packages/ui/src/Button.tsx
danad02: Oh, you're right :smile:

PR #148, opened 2021-01-08
Feature/dialog

### LINKED ISSUE #19: [UI core library] Dialog
https://zeroheight.com/11d0e6dac/p/07395b-dialog
https://zeroheight.com/11d0e6dac/p/19138b-modal

Restyle HzConfirmationModal and HzModal

jankoritak:
Closes #19


packages/ui/__stories__/Dialog.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const affirmation = 'Are you sure, young padawan?'
+const consequences = 'This action may disturb the Force.'
+const onClose = () => console.log('onClose')
+const onAction = () => console.log('onAction')
+const actionLabel = 'Action'
+const disabledTooltip = 'Disabled Tooltip'
+
+const DialogWithPortalFactory: FC<DialogProps> = ({ children, modalClassName, ...props }) => {
+  const id = `s${uuid()}`
```
aigoncharov, line 23: Can't we use 'react-uid'?

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    actionIconLeftAriaLabel,
+    // Rest
+    onAction,
+    ...rest
+  } = restWithActionButton
+
+  return (
+    <ReactModal
+      contentLabel={title}
+      className={cn(styles.modal, className)}
```
aigoncharov, line 103: @jankoritak should we also use `portalClassName` to set dimensions for the portal? Currently, it is unstyled. Because of it, its width is 0px and the overlay cannot stretch.
image

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+      {...rest}
+    >
+      <div className={styles.outline} />
+      <div className={styles.container}>
+        <div data-test="modal-header" className={styles.header}>
+          <div data-test="modal-title" className={styles.title}>
+            {title}
+          </div>
+          {onClose && (
+            <div className={styles.close}>
```
aigoncharov, line 120: Should we add some left margin to it?

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+      onRequestClose={onClose}
+      shouldCloseOnEsc={closable}
+      shouldCloseOnOverlayClick={closable}
+      shouldFocusAfterRender
+      shouldReturnFocusAfterClose
+      {...rest}
+    >
+      <div className={styles.outline} />
+      <div className={styles.container}>
+        <div data-test="modal-header" className={styles.header}>
```
aigoncharov, line 115: Should we use `h` here for semantics?

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+export type ModalActionProps = ModalActionCoreProps & {
+  actionKind?: ButtonProps['kind']
+} & ModalActionDisabledProps &
+  ModalActionIconProps
+
+export type ModalProps = {
+  closable?: boolean
+  children?: ReactNode
+  title: string
+  onClose?: ReactModalProps['onRequestClose']
```
aigoncharov, line 57: I am wondering if we should make it required. What could the case for an un-closeable modal?

packages/ui/src/Button.module.scss
```diff
@@ -128,6 +128,17 @@
     }
   }
 
+  // Kind === danger
+  &.danger {
```
aigoncharov, line 132: @jankoritak could you request some hover, focus and active colors from @pawelklasa ?

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+export type ModalActionProps = ModalActionCoreProps & {
+  actionKind?: ButtonProps['kind']
+} & ModalActionDisabledProps &
+  ModalActionIconProps
+
+export type ModalProps = {
+  closable?: boolean
+  children?: ReactNode
+  title: string
```
aigoncharov, line 56: I am not sure but I think we have cases where we need to support several different actions. Even if we don't I think it is a reasonable expectation.
I would suggest something like this
```suggestion
export type ModalProps = {
  closable?: boolean
  children?: ReactNode
  title: string
  actions: ButtonProps[]
```

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+      overlayClassName={cn(styles.overlay, overlayClassName)}
+      data-test={dataTest}
+      onRequestClose={onClose}
+      shouldCloseOnEsc={closable}
+      shouldCloseOnOverlayClick={closable}
+      shouldFocusAfterRender
+      shouldReturnFocusAfterClose
+      {...rest}
+    >
+      <div className={styles.outline} />
```
aigoncharov, line 113: Why do we need it? AFAICS, the modal itself is not focusable. Only the elements inside.

packages/ui/src/Modal.tsx
jankoritak: I must have broke it during cleanup. 🤦

packages/ui/__stories__/Dialog.stories.tsx
jankoritak: We can. 👍

packages/ui/src/Modal.tsx
jankoritak: Fixed.

packages/ui/src/Modal.tsx
jankoritak: Great idea, thanks for spotting that.

packages/ui/src/Modal.tsx
jankoritak: We actually do have one use-case on MC at the moment. IDP metadata input in SAML provider form. However thinking about it, I don't see why we went this path at all. 

Lemme remove this option.

packages/ui/src/Modal.tsx
jankoritak: It actually is focusable. `react-modal` exposes a prop `shouldFocusAfterRender` to control the behaviour.

This just mirrors out current MC behavior.

packages/ui/src/Modal.tsx
jankoritak: Fair point. Addressed in: https://github.com/hazelcast/frontend-shared/pull/148/commits/b7dca752a32d3db5c02bf509336aed687c0c26e7.

packages/ui/src/Button.module.scss
jankoritak: Requested

jankoritak:
@aigoncharov ready for re-review.

packages/ui/src/Modal.tsx
aigoncharov: @jankoritak AFAIK, according to ARIA guidelines the focus should be on the first focusable element (https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-7), not on the modal itself. I suggest we add `focusTarget: Ref<HTMLELement>` prop to the `Modal` and focus it if provided.
WDYT?

packages/ui/src/Dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+      actionDisabled?: ModalActionProps['disabled']
+      actionDisabledTooltip?: ModalActionProps['disabledTooltip']
+    }
+  | {
+      actionDangerous?: never
+      actionDisabled?: never
+      actionDisabledTooltip?: never
+      actionOnConfirm?: never
+      actionChildren?: never
+    }
```
aigoncharov, line 23: nit: Wouldn;t something simpler work as well?
```suggestion
type DialogActionProps =
  | {
      actionOnConfirm: ModalActionProps['onClick']
      actionChildren: ModalActionProps['children']
      actionDangerous?: boolean
      actionDisabled?: ModalActionProps['disabled']
      actionDisabledTooltip?: ModalActionProps['disabledTooltip']
    }
  | {}
```

packages/ui/src/Dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export type DialogProps = {
+  children?: ReactNode
```
aigoncharov, line 26: What is the use-case for `children`?

packages/ui/src/Modal.tsx
```diff
@@ -0,0 +1,82 @@
+import React, { FC, ReactNode } from 'react'
```
aigoncharov, line 1: Should w also export `setModalAppElement`? http://reactcommunity.org/react-modal/accessibility/#app-element

packages/ui/src/Dialog.tsx
jankoritak: Hmm, we use this pattern at quite some places, so it would be nice to optimise a bit. Using `{}` is rejected by https://github.com/typescript-eslint/typescript-eslint/blob/v4.4.0/packages/eslint-plugin/docs/rules/ban-types.md though.

WDYT about using some type constructor like this one `type NeverObject<T extends object> = T | {[U in keyof T]?: never }`?

packages/ui/src/Dialog.tsx
jankoritak: We support it in `HzConfirmationModal`, but actually, there are no real use-cases. Let me remove that.

packages/ui/src/Dialog.tsx
aigoncharov: I like the idea but I would suggest calling the type something like `WholeOrEmpty`

jankoritak: Makes sense to me. 👍

packages/ui/src/Modal.tsx
aigoncharov: @jankoritak could you point me to any guidelines suggesting to focus the modal itself?

jankoritak: I'm not aware of any. I'm not opposing this idea, I'm rather on board with it. 👍 The reason why I included the property was to mirror original MC behaviour.

>  I suggest we add focusTarget: Ref<HTMLELement> prop to the Modal and focus it if provided.

We can go this path. But I'm wondering whether arbitrary `autoFocus` on given content element is not enough for the job here. WDYT?

aigoncharov: Yep. `autofocus` totally works :)

aigoncharov (approved):
LGTM!

packages/ui/src/Modal.tsx
jankoritak: I'm thinking, what should happen when the modal has no focusable element? That will lead to certain confusion of the user.

Actually, with our Modal/Dialog, that will never happen, as there is always at least `Cancel` button. But according to following source, I think we should implement following.

Source:
- https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-7
> If a dialog contains the final step in a process that is not easily reversible, such as deleting data or completing a financial transaction, it may be advisable to set focus on the least destructive action, especially if undoing the action is difficult or impossible. The Alert Dialog Pattern is often employed in such circumstances.

>If a dialog is limited to interactions that either provide additional information or continue processing, it may be advisable to set focus to the element that is likely to be most frequently used, such as an OK or Continue button.

#### Modal
- In case there are no `actions` provided, we `autoFocus` the `Cancel` button.
- In case there are actions provided and none of them has `autoFocus`, keep the `autoFocus` on `Cancel` button.

#### Dialog
- Note: Built on top of Modal
- In case the action is `primary`, `autoFocus` the action
- In case the action is `danger`, `autoFocus` `Cancel` button

WDYT?

aigoncharov: @jankoritak sounds great! Let's do exactly that! I would also suggest adding `autofocus: false` prop to the modal itself to allow developers set it manually to some item in content if necessary. WDYT?

packages/ui/src/Button.module.scss
jankoritak: The designs have landed in Figma. Will tweak the styles to match it.

packages/ui/src/Modal.tsx
danad02: Oh man, that autofocus sounds really great!

jankoritak: Makes sense. Addressed in https://github.com/hazelcast/frontend-shared/pull/148/commits/139abb63216a52d3673a10fc8f99be8adc0a5e0c.

packages/ui/src/Dialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+  Pick<ModalProps, 'isOpen' | 'title' | 'onClose' | 'parentSelector'>
+
+/*
+ * ### Purpose
+ * Occasionally there's a user-story that's not a part of the main user flow. Such action can be contained in a Dialog.
+ * Dialogs are used to drive simple actions such as a mere confirmation. In case there is a more complex action (e.g larger form, table etc.) use Modal.
+ *
+ * ### General Info
+ * - Dialog is a specification of Modal to fit a particular use-case of confirming an action
+ * - Use affirmation` and `consequences` props to set the content of the Dialog.
```
danad02, line 39: nit:
```suggestion
 * - Dialog is a special kind of Modal. It's made for simple use-cases such as confirmations.
 * - Use affirmation` and `consequences` props to set the content of the Dialog.
```

jankoritak:
FYI guys, I've been talking to Pawel about new designs. They have landed in Figma, so I'm working on the implementation.

danad02 (approved):
Nice! LGTM.

jankoritak:
Update the designs to match latest Figma: https://www.figma.com/file/8mVm6LTbp2Z0RaWWjTZoft/%F0%9F%90%9DHIVE-Hazelcast-Design-System?node-id=20302%3A1. Please, re-review.

packages/ui/__stories__/Dialog.stories.tsx
```diff
@@ -26,38 +25,20 @@ const DialogWithPortalFactory: FC<DialogProps> = ({ modalClassName, ...props })
       <Dialog
         {...props}
         modalClassName={cn(modalClassName, utilsStyles.modal)}
+        portalClassName={utilsStyles.modalPortal}
```
aigoncharov, line 28: @jankoritak shouldn't we add the `portalClassName` for all `Modals`? Why do we have to provide it for every story instead of having the default one in the `Modal` component?

PR #150, opened 2021-01-13
[Core UI] Loader

### LINKED ISSUE #149: [Core UI] Loader
Use FadeLoader from https://www.davidhu.io/react-spinners/. Should be available in two colors: `primary` and `contrast` (for dark backgrounds).

aigoncharov:
- Add `Loader` component
- Add `loading` prop to `Button`

Closes #149


jankoritak (approved):
LGTM! Nice job with the custom animation. :)

packages/ui/src/Loader.tsx
```diff
@@ -0,0 +1,51 @@
...
+import styles from './Loader.module.scss'
+
+export type LoaderProps = {
+  className?: string
+  size?: IconSize
+  kind?: 'primary' | 'contrast' | 'inherit'
+  children?: never
+} & Pick<SVGAttributes<SVGElement>, 'aria-hidden'>
+
+export const Loader: FC<LoaderProps> = ({ className, size = 'normal', kind = 'inherit', ...props }) => (
```
jvorcak, line 15: Why don't we just `import { FadeLoader } from 'react-spinners` and use it, but we embed the SVG and do all the animation manually inside SVG? Is there any special reason?

jvorcak (changes requested):

packages/ui/src/Loader.tsx
aigoncharov: @jvorcak I left a comment on why I imported the SVG
```
// https://github.com/feathericons/feather/blob/a718a7e9c39447202f703783336e8ba1c8e32405/icons/loader.svg#L1
// We cannot use it directly as it has `line` elements not in a sequential order
// And we need the sequential order to easily calculate animation delay
```

In essence, it is not the same SVG. I reordered `line` elements.

I did not want to add yet another dependency. I felt like we could save some kB animating a simple SVG.

jvorcak (approved):

PR #151, opened 2021-01-13
WIP Feat/multiselect

utkukaratas:
hi folks, this is still WIP but i'd be glad if you check it out 

instead of a seperate MultiSelect component, the SelectField is modified:

- .isMulti prop is introduced.
- SelectFieldValue type is introduced which will need to be used wherever the .value prop is handled. (check out the story changes)

so WDYT? if you guys think this is complex we could go the separate component way. not married to the idea as usual. 

cc @aigoncharov @jankoritak @danad02


packages/ui/src/SelectField.tsx
```diff
@@ -36,22 +37,27 @@ export type SelectFieldOption<V = string> = {
...
   error?: string
+  isMulti?: boolean
 }
+
 export type SelectFieldCoreDynamicProps<V> =
   | {
       isClearable: true
-      value: SelectFieldOption<V> | null
-      onChange: (newValue: SelectFieldOption<V> | null) => void
+      value: SelectFieldValue<V> | null
```
aigoncharov, line 52: @utkukaratas I am not sure about this typing. Potentially, it allows users to pass value: V and the onChange that expects V[] back. It was one of the reasons I suggested to have MultiSelect as a separate component

packages/ui/src/SelectField.tsx
```diff
@@ -103,6 +110,27 @@ export const SelectField = <V,>({
...
   }, [menuPortalTarget])
 
+  // Custom styling
+  //
+  // This is the recommend way of styling react-select:
+  // (https://react-select.com/styles#using-classnames)
+  //
+  // ps. Also it's impossible to style the focused state of
+  // multiValue tags via CSS.
+  const customStyles = {
```
aigoncharov, line 120: @utkukaratas still react-select allows us to use classes. Should we stick to classes and SCSS for consistency?

aigoncharov: I think @jankoritak did a great job before with Select and demonstrated how it could be easily done with simple class overrides, or component replacements if necessary. WDYT?

packages/ui/src/SelectField.tsx
utkukaratas: hmm well if a type-happy coder like you finds it confusing we better seperate it :)) incoming...

packages/ui/src/SelectField.tsx
utkukaratas: @aigoncharov css was my first attempt too, but it's impossible to style the focused state of multiValue tags as per our hover-focus designs. i'll show you what i mean with an ss...

packages/ui/src/SelectField.tsx
aigoncharov: Let's ask the team.

packages/ui/src/SelectField.tsx
aigoncharov: If we cannot style the existing component for dinner reason, couldn't we replace it with our own one?
https://react-select.com/props#components

packages/ui/src/SelectField.tsx
utkukaratas: (context: do we create a seperate MultiSelect component **VS** augment SelectField with more intricate types to handle multiple selections)

jankoritak: The current type seems good enough for common use-cases, so I'd be fine with keeping that. I don't have a strong preference, though.

packages/ui/src/SelectField.tsx
danad02: I'd like to vote for replacement it even if the styling is possible. I find it cleaner.

packages/ui/src/SelectField.tsx
danad02: Making almost identical component again because of this seems like an overkill to me. I know we want to be clean but maybe that's the case we're fine with. A comment might help if the user struggles with types.

danad02: @utkukaratas @aigoncharov Maybe this is a solution?
```
export type SelectFieldCoreDynamicProps<V> =
  | {
      isClearable: true
      value: SelectFieldOption<V> | null
      onChange: (newValue: SelectFieldOption<V> | null) => void
    }
  | {
      isClearable?: false
      value: SelectFieldOption<V>
      onChange: (newValue: SelectFieldOption<V>) => void
    }
  | {
      isClearable: true
      value: SelectFieldOption<V>[] | null
      onChange: (newValue: SelectFieldOption<V>[] | null) => void
    }
  | {
      isClearable?: false
      value: SelectFieldOption<V>[]
      onChange: (newValue: SelectFieldOption<V>[]) => void
    }
```

aigoncharov: @danad02 good idea!

utkukaratas: that's a mouthful but why not. thx @danad02. on its way...

utkukaratas:
hi there @jankoritak,

we have an issue with **SelectField / "menuIsOpen=true" stories** when **onChange/setValue combo** is used on the Storyboard (see my latest commit pls). if you try to select something, you'll see the popup jump around. do you have any tips why this could happen. i better solve this while i'm on it.

utkukaratas:
verify-all

utkukaratas:
verify

packages/ui/src/SelectField.tsx
```diff
@@ -40,18 +46,35 @@ export type SelectFieldCoreStaticProps = {
   name: string
   onBlur?: (e: FocusEvent<HTMLElement>) => void
   error?: string
+  isMulti?: boolean
```
danad02, line 49: Since `isMulti` is present in `SelectFieldCoreDynamicProps` we can remove it here.

packages/ui/__tests__/SelectField.test.tsx
```diff
@@ -210,4 +210,45 @@ describe('SelectField', () => {
...
+
+  it('Renders multiple selections with correct props', async () => {
+    const onChange = jest.fn()
+
+    const value = [selectOptions[0], selectOptions[1]]
+    const wrapper = await mountAndCheckA11Y(
+      <SelectField name={selectName} label={selectLabel} isMulti={true} options={selectOptions} value={value} onChange={onChange} />,
+      {
+        // xxx axeOptions: SELECT_FIELD_AXE_OPTIONS,
+      },
```
danad02, line 222: Do we need this?

danad02, line 222: Do we need this?

utkukaratas: it seems it's gone from the master too, removing...

packages/ui/src/SelectField.tsx
utkukaratas: nice catch, thx!

utkukaratas:
btw @jankoritak, i'll wait for your approval for the issue above before merging.

packages/ui/src/SelectField.tsx
jankoritak: It seems like this one was not addressed yet. Can you look into it, @utkukaratas? Once this is addressed, it seems good for approval.

utkukaratas: @jankoritak this was an outdated convo. resolving it...

jankoritak: Got it. Could you rebase the branch? AFAIC, the PR is good to go.

jankoritak (approved):

packages/ui/src/SelectField.tsx
utkukaratas: done @jankoritak

danad02 (approved):
LGTM. Nice job :+1:

packages/ui/src/SelectField.module.scss
```diff
@@ -95,6 +95,19 @@ $select: ':global .hz-select-field__control';
   }
 }
 
+// Styles for MultiValue chips - hover and focused states.
+// We inject `.multiValueIsFocused` from js (it does not have a :focused state as its tabIndex is not given).
+.multiValueIsFocused {
+  :global .hz-select-field__multi-value__remove {
+    background-color: c.$colorWarning;
+    border: 1px solid c.$colorAccessibilityOutline;
```
aigoncharov, line 103:
```suggestion
    border: c.$borderWidth solid c.$colorAccessibilityOutline;
```

packages/ui/src/SelectField.module.scss
```diff
@@ -95,6 +95,19 @@ $select: ':global .hz-select-field__control';
...
+  }
+}
+:global .hz-select-field__multi-value__remove:hover {
+  background-color: c.$colorWarning;
+  border: 1px solid c.$colorPrimary;
```
aigoncharov, line 108:
```suggestion
  border: c.$borderWidth solid c.$colorPrimary;
```

aigoncharov: Should we also add a `1px` transparent border for the unhovered state, so the cell does not jump when hovered?

packages/ui/__stories__/SelectField.stories.tsx
```diff
@@ -300,6 +301,56 @@ export const NonSearchableOpen = () => {
...
+      onBlur={() => logger.log('blur')}
+      onChange={setValue}
+    />
+  )
+}
+
+export const MultipleSelectionsMultipleRows = () => {
+  const [currentValue, setValue] = useState([...options])
+  return (
+    <SelectField
```
aigoncharov, line 322: Should we wrap it with a `div` with some max-wdith, say, 400px, to make it really multiple rows? On the current screenshot it is only one row

packages/ui/src/SelectField.tsx
```diff
@@ -31,6 +31,12 @@ const Input = (props: any) => {
   return <components.Input {...props} autoComplete="chrome-off" />
 }
 
+// this component is implemented because we want to style things with the css (as opposed to the react-select recommended way of styling things via js).
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const MultiValue = (props: MultiValueProps<any>) => {
+  return <components.MultiValue {...props} className={cn({ [styles.multiValueIsFocused]: props.isFocused })} />
```
aigoncharov, line 37: @utkukaratas besides forcused state of the chip, should we style its default state? I would suggest the following:
- Use one of our neutral colors for its backgorund
- Use typography helpers (typographyBodySmall?) to use standard font sizes
- Use colorText for the text color
- Replace remove icon with the one from react-feather, so it matches what we use in our `Toast` component.
- Optionnally, try matching padding and margins with what we have in our `Toast`component

WDYT?

packages/ui/src/SelectField.tsx
```diff
@@ -41,17 +47,33 @@ export type SelectFieldCoreStaticProps = {
...
       value: SelectFieldOption<V> | null
       onChange: (newValue: SelectFieldOption<V> | null) => void
     }
   | {
       isClearable?: false
+      isMulti?: false
       value: SelectFieldOption<V>
       onChange: (newValue: SelectFieldOption<V>) => void
     }
+  | {
```
aigoncharov, line 64: Should we also use the same technique in `SelectFieldFormikProps`? Could we also have a story for the formik integration of MultiSelect?

aigoncharov (changes requested):
@utkukaratas left a couple of comments! Great job so far!

packages/ui/src/SelectField.tsx
utkukaratas: we now use our IconButton for the button. resolving...

packages/ui/src/SelectField.tsx
utkukaratas: this is done and committed. i'll request a review specifically for this.

danad02 (approved):

utkukaratas:
hi @aigoncharov this is ready for a review. all of the issues above are handled.

aigoncharov:
@utkukaratas I have tried the storybook. I cannot remove selected items by clicking their `x`. The click opens the dropdown menu instead.l

packages/ui/src/SelectField.tsx
aigoncharov: @utkukaratas should we resolve all other proposals? I see that only the remove icon was changed.
I still suggest we do:
- Use one of our neutral colors for its backgorund
- Use typography helpers (typographyBodySmall?) to use standard font sizes
- Use colorText for the text color
- Try matching padding and margins with what we have in our Toastcomponent

WDYT?

aigoncharov:
@utkukaratas could you also take a look at https://github.com/hazelcast/frontend-shared/pull/151#discussion_r568357866 ?

aigoncharov (changes requested):
@utkukaratas I think we are getting really close! Left a couple of comments

packages/ui/src/SelectField.tsx
utkukaratas: @aigoncharov i've used our IconButton instead of doing those manually. much cleaner IMO. it seems recent IconButton changes broke this PR but now that's fixed too.

packages/ui/src/SelectField.tsx
```diff
@@ -31,6 +31,41 @@ const Input = (props: any) => {
...
+}
+
+// Self styled version of the MultiValue/Remove button
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const MultiValueRemove = (props: any) => {
+  // We have to pass down the onClick and onMouseDown handler to our IconButton
+  // otherwise a click is a noop.
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,  @typescript-eslint/no-unsafe-assignment
+  const onClick: React.MouseEventHandler = props.innerProps.onClick
```
aigoncharov, line 51: Can't we stop propagation for the onClick and not introduce the mouseDown handler?

packages/ui/src/SelectField.tsx
aigoncharov: @utkukaratas I see that you used IconButton before. I am talking abou the color, padding and typography of the chip that represents a selected item. Currently, the chip does not match our styling.

packages/ui/src/SelectField.tsx
utkukaratas: nope, that doesn't work. it was the MouseDown chain of events that should be stopped.

packages/ui/src/SelectField.tsx
utkukaratas: ah! i misunderstood that. will handle it...

utkukaratas: @aigoncharov 

- all of the issues above should be resolved. 
- it should be pixel perfect with the new design. (except some parts in the design those are not multiples of 4px)

utkukaratas:
verify

utkukaratas:
pr-builder seems to be stuck :/

utkukaratas:
verify

aigoncharov:
@utkukaratas it looks amazing! Great job! On last bit
image
Should we match the paddings for the placeholder text for an empty input?

packages/ui/src/SelectField.module.scss
```diff
@@ -95,6 +95,47 @@ $select: ':global .hz-select-field__control';
...
+    align-items: flex-start;
+    padding-top: c.$grid;
+  }
+
+  .multiValue {
+    background-color: c.$colorNeutralWhite !important;
+    border: c.$borderWidth solid c.$colorNeutralLight;
+    border-radius: c.$borderRadius;
+    margin: c.$grid * 0.5 !important; // margin collapsing is not there in flexbox. we give half margins to multiValues.
+    height: c.$grid * 5.5; // makes 22px; design is 21px.
```
aigoncharov, line 114: Good thinking! I totally agree with making it a sane increment of `grid` instead of being 100% pixel perfect

packages/ui/src/SelectField.module.scss
```diff
@@ -95,6 +95,47 @@ $select: ':global .hz-select-field__control';
...
+  .multiValueLabel {
+    @include c.typographyBodySmall;
+
+    color: c.$colorText;
+    padding: 0 c.$grid;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    box-sizing: border-box;
+  }
```
aigoncharov, line 136: nit
```suggestion
  .multiValueLabel {
    @include c.typographyBodySmall;
    @inclide h.overflowEllipsis;

    color: c.$colorText;
    padding: 0 c.$grid;
    box-sizing: border-box;
  }
```

packages/ui/src/SelectField.tsx
```diff
@@ -31,6 +31,50 @@ const Input = (props: any) => {
   return <components.Input {...props} autoComplete="chrome-off" />
 }
 
+// This component is implemented because we want to style things with the css (as opposed to the react-select
+// recommended way of styling things via js).
+//
+// We simply inject `.multiValueIsFocused`.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const MultiValue = (props: MultiValueProps<any>) => {
```
aigoncharov, line 39: nit: SHould `MultiValue` come after `MultiValueRemove`and `MultiValueLabel` since it references them? AFAIU, `consts` are not hoisted all the way to the top, so it works only because these arrow functions are converted to regular functions during transpilation

aigoncharov:
@utkukaratas clear all button is kind of misaligned. Could you take a look?
image

packages/ui/src/SelectField.tsx
utkukaratas: i don't get the point here??? MultiValue uses the other 2 components inside its body - so they will be referenced at the time of the call. declaration order should not be an issue here in any case.

packages/ui/src/SelectField.module.scss
utkukaratas: there're also some 1px mismatches between paddings and what not. (i'll demo this to @pawelklasa for his blessings to the implementation :D designer folks have ultra precise / lazer zoom eyes for these kind of things.)

packages/ui/src/SelectField.tsx
aigoncharov: @utkukaratas here is a simple reproducer https://www.typescriptlang.org/play?noUnusedLocals=true&noUnusedParameters=true&declaration=false&target=99#code/MYewdgzgLgBAhjAvDAFASiQPhgbwFAwwBG6eAvnnKXqJLEUqhotrRCADYCmAdByAHMUAcgAWAW0nC0QA
Try running it and see the console

aigoncharov: We get away with it because webpack transpiles it to https://www.typescriptlang.org/play?noUnusedLocals=true&noUnusedParameters=true&declaration=false&target=99#code/PTAEGMHsDsGcBdQENQF5QAoCUaB8oBvAKBFDICNsSwBfa5K+qORctTHVfZ2SAGwCmAOj6QA5hgDkACwC28yViL0AKgCckcAA4BLQQBNQ5AJ6gA7gPJak4ANah4kZQDMArtHDwdMZB0JEKKjokRjcPLx9KHGIyHn5hUQkZBSUaIA

utkukaratas: ok calling a() before declaring "b" might of course be problematic but in our case noone is calling MultiValue before declaring the other two. and MultiValue is instantiated much much later, in fact ages later, when SelectField is instantiated somewhere in the app code.

aigoncharov: @utkukaratas yeah, you are right. Here is a great article that helped me to brush up my knowledge about lexical scoping BTW https://blog.bitsrc.io/hoisting-in-modern-javascript-let-const-and-var-b290405adfda

utkukaratas: @aigoncharov you're a review machine dude, and every machine needs a reboot at one point, you just got segfaulted :D

utkukaratas:
setting the PR to draft. that placeholder margin issue opened such a can of worms :(((

aigoncharov:
> setting the PR to draft. that placeholder margin issue opened such a can of worms :(((

Can;t we just add extra margin to the `.hz-select-field__placeholder` element when it is a child of a multi value select?

utkukaratas:
> Can;t we just add extra margin to the `.hz-select-field__placeholder` element when it is a child of a multi value select?

yes but thet the input (blinking caret) container looks off. tried a few different ways. that resulted lots of cascading failures :/ i'll sleep on this to find a better way.

aigoncharov:
@utkukaratas when container has a value it has `.hz-select-field__value-container--has-value` class applied. You can change the horizontal margin here https://github.com/hazelcast/frontend-shared/pull/151/files#diff-bc420743ebb0252dfa1d6d8e7cdfd29a04c885dbc6127186fe1a98d68537a83dR113 only when there is a value/values

utkukaratas:
> @utkukaratas when container has a value it has `.hz-select-field__value-container--has-value` class applied. You can change the horizontal margin here https://github.com/hazelcast/frontend-shared/pull/151/files#diff-bc420743ebb0252dfa1d6d8e7cdfd29a04c885dbc6127186fe1a98d68537a83dR113 only when there is a value/values

thx, this has been a great trick! though that css is now a living organism on its own :/

i'll demo this to @pawelklasa tomorrow and then will submit.

utkukaratas:
@aigoncharov got the approval from pawel; ready for code review.

packages/ui/src/SelectField.tsx
```diff
@@ -31,6 +31,53 @@ const Input = (props: any) => {
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const MultiValueRemove = (props: any) => {
+  // We have to pass down the onClick
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,  @typescript-eslint/no-unsafe-assignment
+  const onClick: React.MouseEventHandler = props.innerProps.onClick
+
+  const handleMouseDown = (event: React.MouseEvent) => {
+    event.preventDefault()
+    event.stopPropagation()
+    event.button === 0 && onClick && onClick(event)
```
aigoncharov, line 71: @utkukaratas why do we need this line?

aigoncharov: Can't we just pass `onClick` to the button? BTW, should we just use `IconButton`?

aigoncharov: Oh, I see now that it is custom styled.The question regarding `onClick` still stands.

aigoncharov: @utkukaratas there is a small visual problem 
image

aigoncharov: As you can see the hover background for the remove button overlaps with the focus shadow. Shouldn't it go underneath?

aigoncharov: We'll probably need a transparent absolutely positioned element on top of the button to draw the shadow on the very top

utkukaratas: > @utkukaratas why do we need this line?

we need to pass the onclick given from the parent to our custom div. it's done in `mousedown`  to do everything in one fell swoop and keep the linter happy.

aigoncharov: @utkukaratas personally, I would expect the removal to happen only when I release the mouse, but it is a nit. It is just that click is usually happens after mouse is up, so currently it feels a bit unnatural. WDYT? What error does a linter give you?

aigoncharov: I tried it. It is a keybord listener error, right? I would suggest to just put a eslint ignore on this line.  WDYT?

utkukaratas: > We'll probably need a transparent absolutely positioned element on top of the button to draw the shadow on the very top

just modified the box shadow to be NOT inset ;)

aigoncharov (approved):
Great job! let's merge this one in!

PR #159, opened 2021-01-18
feat(PasswordFieldn): add support for "id" prop

jozefhruska:


packages/ui/src/PasswordField.tsx
```diff
@@ -22,6 +22,7 @@ import styles from './PasswordField.module.scss'
 
 type PasswordFieldCoreProps = {
   name: string
+  id?: string
```
aigoncharov, line 25: `id` is automatically generated by the `TextField`. You'll have to propagate it there as well.
BTW, should it go on L42 instead?

packages/ui/src/PasswordFieldFormik.tsx
```diff
@@ -7,6 +7,7 @@ import { ExtractKeysOfValueType } from './utils/types'
 
 export type PasswordFieldFormikProps<V extends object> = PasswordFieldExtraProps & {
   name: ExtractKeysOfValueType<V, string | undefined>
+  id?: string
```
aigoncharov, line 10: Is it really passed by formik or is it just an extra prop?

jozefhruska: Not entirely sure what do you mean. Could you elaborate please?

packages/ui/src/PasswordField.tsx
```diff
@@ -38,7 +38,7 @@ export type PasswordFieldExtraProps = {
   withIcon?: boolean
   visibleRef?: Ref<VisibleRef>
 } & DataTestProp &
-  Pick<InputHTMLAttributes<HTMLInputElement>, 'autoFocus' | 'disabled' | 'autoComplete' | 'required' | 'placeholder'>
+  Pick<InputHTMLAttributes<HTMLInputElement>, 'id' | 'autoFocus' | 'disabled' | 'autoComplete' | 'required' | 'placeholder'>
```
aigoncharov, line 41: That `id` is passed to `TextField` that generates its own `id`. Should we use the external `id` that we pass (if present) instead of the autogenerated one?

jozefhruska: Well, the autogenerated `id` is overloaded by `{...htmlAttrs}`, so it works as expected.

jozefhruska: But maybe that's not entirely clear on first look. Maybe we can do it like this in `TextField`?
```
id={htmlAttrs.id ?? autoId}
```
What do you think?

aigoncharov: @jozefhruska the autogenerated id is overridden only for the `input`, but not for `Help`, `Error` and `Label`

jozefhruska: Good catch! I added an overload to `TextField`. Sorry, this is a quick change for Cloud release... Guess I rushed things a bit 🙈

aigoncharov (approved):

danad02 (approved):

danad02 (approved):

PR #163, opened 2021-01-19
Bugfix/icon dimensions

aigoncharov:
- Fix `Icon` size. We cannot set size in rems directly. SVG does not support `rem` in Firefox - https://bugzilla.mozilla.org/show_bug.cgi?id=1231147
- Fix setAppModal export. For some reason it does not work as a direct re-exporet in MC tests.
- SImplify IconButton


danad02 (approved):
LGTM

PR #164, opened 2021-01-19
Relax axe rules

aigoncharov:


danad02 (approved):
Just wanted to suggest this change because I had to override axe in most of my Table tests. Thanks!

PR #165, opened 2021-01-19
Update wrapper after act

aigoncharov:
Potenatially, something could have changed the layout inside of act (like an effect). We need to update the wrapper to return the updated component.


jankoritak (approved):

danad02 (approved):
It's true that we have a lot of these initial acts and updates in MC tests, makes sense to do the same thing here. Nice :+1:

PR #170, opened 2021-01-21
Pass properties to Link in Alert component

jvorcak:


packages/ui/__tests__/Alert.test.tsx
```diff
@@ -97,19 +120,19 @@ describe('Alert', () => {
...
 
-  const alertActionsTestData = [
+  const alertActionsTestData: Array<[string, Array<AlertActionLink | AlertActionButton>]> = [
     ['1 action', [AlertAction1]],
     ['2 actions', [AlertAction1, AlertAction2]],
   ]
 
-  it.each(alertActionsTestData)('Renders Alert with %s', async (_, actions) => {
-    const wrapper = await mountAndCheckA11Y(<Alert type="success" title={title} content={content} closeToast={noOp} />)
+  it.each(alertActionsTestData)('Renders Alert with %s', async (x, actions) => {
```
jankoritak, line 128: Nit: Just because of the convetion.

```suggestion
  it.each(alertActionsTestData)('Renders Alert with %s', async ( _, actions) => {
```

packages/ui/__tests__/Alert.test.tsx
```diff
@@ -97,19 +120,19 @@ describe('Alert', () => {
...
+    const wrapper = await mountAndCheckA11Y(<Alert type="success" title={title} content={content} closeToast={noOp} actions={actions} />)
 
     wrapper
       .findDataTest('alert-actions')
       .children()
       .forEach((child, cI) => {
-        expect(child.props()).toMatchObject(actions[cI])
+        expect(child.props()).toHaveProperty('children', actions[cI].text)
```
jankoritak, line 135: I wonder, why fall back to just the text here? Does `expect(child.props()).toMatchObject(actions[cI])` break?

jankoritak (commented):
Looks good, leaving 2 questions.

packages/ui/__tests__/Alert.test.tsx
jvorcak: The problem is that there is a props manipulation inside the Alert. e.g. text -> children. Props are just named differently. So I tried to test manually, but then we have a problem that we have 2 types of Actions (Link & Button), and undefined are not handled in `toMatchObject`.

So I've decided to make this test only to make sure we render it (as the name suggests) and handle exact props matching in a separate tests ☝️

jvorcak: Note for others: The problem with the original test is that actions are not passed and `toMatchObject` silently ignores empty object.

jankoritak: I see, I would alter (specify) the name of the test to match the semantics then. Does that make sense?

jankoritak (approved):

PR #173, opened 2021-01-25
fix: icon/iconbutton outline spacing

### LINKED ISSUE #172: [Core UI] Enhance IconButton alignment
See https://github.com/hazelcast/management-center/pull/4172#discussion_r561902594

jankoritak:
Fixes #172


danad02 (approved):
LGTM

aigoncharov:
Should we keep icons inline/inline-block/inline-flex to easily insert them in any text?

jankoritak:
> Should we keep icons inline/inline-block/inline-flex to easily insert them in any text?

That's a great point. Addressed.

@danad02, could you please run the visual tests for this PR? 🙏

danad02:
@jankoritak There's a small regression caused by the Icon's <div> element addition:
image
The icons in TextField do not change color when in hovered/focused/error state.

jankoritak:
Wow, that's a great catch! Will look into it, thanks!

PR #176, opened 2021-02-11
WIP - Code component

utkukaratas:
```
  Our Code component is based on react-syntax-highlighter:
  https://github.com/react-syntax-highlighter/react-syntax-highlighter#readme
  * react-syntax-highlighter itself is based on either one of these two popular js highlighters: Prism.js, highlightjs.org.
  * We're using the light version of the Prism backend.
  * Only a select number of languages declared above are supported to keep it light.
  * Using other languages and themes is possible with .language and .theme props. See storybook for an example.
```


utkukaratas:
verify

packages/ui/src/Code.tsx
```diff
@@ -0,0 +1,94 @@
...
+import kotlin from 'react-syntax-highlighter/dist/esm/languages/prism/kotlin'
+import python from 'react-syntax-highlighter/dist/esm/languages/prism/python'
+import scala from 'react-syntax-highlighter/dist/esm/languages/prism/scala'
+
+// Themes
+import vs from 'react-syntax-highlighter/dist/esm/styles/prism/vs'
+import darcula from 'react-syntax-highlighter/dist/esm/styles/prism/darcula'
+
+// Register these builtin languages we'd use frequently
+SyntaxHighlighter.registerLanguage('jsx', jsx)
```
aigoncharov, line 26: I am wondering if we need to setup default languages at all. Every team might have different needs. I would suggest we do not bundle any languages, declare `react-syntax-highlighter` a peer dep and let every team pick their own languages. WDYT?

aigoncharov:
@utkukaratas AFAICS, the component only highlights a static string, i.e. we cannot use it as an input to edit code. We need that functionality for `Scripting` in MC. Am I missing something?

utkukaratas:
> @utkukaratas AFAICS, the component only highlights a static string, i.e. we cannot use it as an input to edit code. We need that functionality for Scripting in MC. Am I missing something?

@aigoncharov yes, this component is basically a nicer <pre /> 

code-mirror must be used if input is needed (which we are using in MC code).

aigoncharov:
@utkukaratas should we try https://github.com/antonmedv/codejar? Looks lightweibght and promising

utkukaratas:
> @utkukaratas should we try https://github.com/antonmedv/codejar? Looks lightweibght and promising

this also seems to depend on Prismjs just like the current solution. if we ever rewrite the scripting screens this might be something we could try to get rid of codemirror. (but also codemirror seems to be rewritten recently (v6); we should also look at that before we think of cutting it out.)

utkukaratas:
@aigoncharov okay then.. i had assumed code component is only needed for readonly purposes (code blocks in blogposts/docs etc..) and frontend-shared would not contain an editor at all. now the requirements changed, i'll be on codemirror path. WIP...

danad02:
> In fact, I know for sure that we need the code editor, and I am not sure if we need to display readonly code in MC at all.

We do have a few places where we want to show readonly XML or YAML.

utkukaratas:
closing this; there'll be a codemirror based version in the works...

PR #178, opened 2021-02-12
fix: react-use imports and remove use-deep-compare

jankoritak:
Fixes the IE11 (storybook) crashes

- Corrects imports of `react-use`
- Replaces `use-deep-compare/useDeepCompareMemo` with custom `useDeepCompareMemo` fn.


danad02 (approved):
Great finds! :+1:

danad02:
Btw did `use-deep-compare` also cause IE11 trouble?

jankoritak:
Yes, that was the issue introduced by `table` PR.

PR #179, opened 2021-02-12
Feat: add optional heading content to Card

danad02:
Add optional `headingContent` to the Card. It can be anything.


aigoncharov:
@danad02 could we simplify it a bit and just make `title` prop accept any JSX?

danad02:
@aigoncharov Good idea. How come this idea didn't strike me right away is beyond me. I'm tired probably :smile:

danad02:
@aigoncharov I think that in this case we can also just remove the left side icon. WDYT?

aigoncharov:
> @aigoncharov I think that in this case we can also just remove the left side icon. WDYT?

I am wondering where is the line where we have too much freedom. Probably, even making the title any arbitrary JSX might be too much as it could allow us breaking the design guidelines. What if wekeep the title as a string, but add `headerLeftSlot` prop which would be any JSX. Then we would be able to make sure there are proper padding between the elements. WDYT?

danad02:
@aigoncharov I added it a `headingContent` prop to the Card component. Check it out: https://github.com/hazelcast/frontend-shared/pull/179/commits/6f6646b3ba7da789e2627f650d31eb61c3c6ba0e

aigoncharov (approved):
LGTM!

PR #185, opened 2021-02-24
Feat: add isCreatable option to SelectField

danad02:
Needed for some cases in MC.


packages/ui/src/SelectField.tsx
```diff
@@ -201,30 +204,58 @@ export const SelectField = <V,>({
...
-          components={{
-            DropdownIndicator,
-            ClearIndicator,
-            Input,
-            MultiValue,
-          }}
-          {...rest}
-        />
+        {isCreatable ? (
+          <ReactSelectCreatable<SelectFieldOption<V>>
```
aigoncharov, line 208: nit: They have the same props, right? Should we then do something like this?
```tsx
const SelectComponent = isCreatable ? ReactSelectCreatable : ReactSelect
...
return <SelectComponent  .../>
```

aigoncharov (approved):

packages/ui/src/SelectField.tsx
danad02: Creatable has the same props as classic select and **some additional** props. I ended up not using any of the extra props, so for now we can change the implementation to look as you suggest.

aigoncharov: @danad02 alternatively we could create a single object with all props and apply it to both components. It is probably a better approach.

aigoncharov:
```tsx
const props = {}
...
return isCreatable ? <CreatableSelect {...props} /> : <Select {...props} />
```

PR #193, opened 2021-03-02
Feature/react select single value

jvorcak:
This PR allows our ReactSelect to work with values instead of objects, based on the options property.


packages/ui/src/SelectField.tsx
```diff
@@ -181,6 +182,37 @@ export const SelectField = <V,>({
...
+    (option: SelectFieldOption<V | null>) => {
+      ;(onChange as (newValue: V | null) => void)(option === null ? null : option.value)
+    },
+    [onChange],
+  )
+
+  const selectedOption: SelectFieldOption<V> | SelectFieldOption<V>[] | undefined = useMemo(() => {
+    if (isMulti) {
+      // if it's multi value, let's transform a value array to an array containing SelectFieldOptions
+      const valueArray = value as V[] | null
```
jankoritak, line 202: Suggestion: WDYT about providing a type-guard here instead of inline `as` cast?

packages/ui/src/SelectField.tsx
```diff
@@ -181,6 +182,37 @@ export const SelectField = <V,>({
...
+      return (
+        valueArray?.map((val) => ({
+          value: val,
+          // we need to find instead of value: object mapping so that we can support objects as values
+          label: (options.find(({ value }) => value === val) as SelectFieldOption<V>).label,
+        })) ?? []
+      )
```
jankoritak, line 209: Nit: We can remove these here. 

```suggestion
      return
        valueArray?.map((val) => ({
          value: val,
          // we need to find instead of value: object mapping so that we can support objects as values
          label: (options.find(({ value }) => value === val) as SelectFieldOption<V>).label,
        })) ?? []
```

jankoritak (approved):
LGTM, leaving 2 suggestions.

packages/ui/src/SelectField.tsx
jvorcak: I'm working on adjusting this to a creatable option. I'll push it soon, can you please re-review afterwards?

jankoritak: Sure, sry. Missed the `WIP`.

packages/ui/src/SelectField.tsx
```diff
@@ -106,28 +106,45 @@ export type SelectFieldCoreDynamicProps<V> =
...
   | {
       isClearable?: false
       isMulti: true
-      value: SelectFieldOption<V>[]
-      onChange: (newValue: SelectFieldOption<V>[]) => void
+      value: V[]
+      onChange: (newValue: V[]) => void
     }
 
+// Since the user input is string, let's allow creatable only for string
```
aigoncharov, line 131: Good thinking!

packages/ui/src/SelectField.tsx
```diff
@@ -154,6 +174,43 @@ const getMenuContainer = (menuPortalTarget: 'body' | 'self' | HTMLElement | null
...
+ * @param isMulti - boolean indicating whether we're in a multiple mode or not
+ */
+export function getSelectedOptionFromValue<V>({ value, options, isMulti }: GetSelectedOptionFromValueProps<V>) {
+  if (isMultipleModeGuard(value, isMulti)) {
+    // if it's multi value, let's transform a value array to an array containing SelectFieldOptions
+    return (
+      value?.map((val) => ({
+        value: val,
+        // we need to find instead of value: object mapping so that we can support objects as values
+        label: (options.find(({ value }) => value === val) as SelectFieldOption<V>)?.label ?? val,
```
aigoncharov, line 196: nit: Should we a hashmap out of `options` for O(1) access?

packages/ui/src/SelectField.tsx
```diff
@@ -154,6 +174,43 @@ const getMenuContainer = (menuPortalTarget: 'body' | 'self' | HTMLElement | null
...
+  isMulti: boolean
+}
+
+/**
+ * Transforms the value passed to our Select component so that it's in a format as underlying react-select expects
+ * @param value - single value or an array of values
+ * @param options - an array of objects describing all possible options with their labels
+ * @param isMulti - boolean indicating whether we're in a multiple mode or not
+ */
+export function getSelectedOptionFromValue<V>({ value, options, isMulti }: GetSelectedOptionFromValueProps<V>) {
```
aigoncharov, line 189: nit: Should it ba a const arrow function?

packages/ui/src/SelectField.tsx
```diff
@@ -184,6 +242,35 @@ export const SelectField = <V,>({
...
+  )
+
+  const onChangeFn = React.useCallback(
+    (option: SelectFieldOption<V | null>) => {
+      ;(onChange as (newValue: V | null) => void)(option === null ? null : option.value)
+    },
+    [onChange],
+  )
+
+  const onCreateOption = React.useCallback(
```
aigoncharov, line 259: Why do we need this one? AFAIU, react-select handles this logic automatically. I tried to remove it locally and it just worked. Then we'll also be able to remove `isCreatableGuard
`. Moreover, with the current implementation, a single-value creatable select does not work.

aigoncharov (commented):
@jvorcak great job and wonderful idea! Left a single comment to address. After that - good to go.

packages/ui/src/SelectField.tsx
```diff
@@ -154,6 +174,43 @@ const getMenuContainer = (menuPortalTarget: 'body' | 'self' | HTMLElement | null
...
+ * Transforms the value passed to our Select component so that it's in a format as underlying react-select expects
+ * @param value - single value or an array of values
+ * @param options - an array of objects describing all possible options with their labels
+ * @param isMulti - boolean indicating whether we're in a multiple mode or not
+ */
+export function getSelectedOptionFromValue<V>({ value, options, isMulti }: GetSelectedOptionFromValueProps<V>) {
+  if (isMultipleModeGuard(value, isMulti)) {
+    // if it's multi value, let's transform a value array to an array containing SelectFieldOptions
+    return (
+      value?.map((val) => ({
```
danad02, line 193: I feel like that guard is not entirely correct. We tell TypeScript that value is certainly `V[]` while in fact it could also be `null`. Then we check with optional chaining if value truly is not null. If you remove the `?` from `value?.map()` TypeScript won't error while in fact it should.

packages/ui/src/SelectField.tsx
```diff
@@ -154,6 +174,43 @@ const getMenuContainer = (menuPortalTarget: 'body' | 'self' | HTMLElement | null
...
+        value: val,
+        // we need to find instead of value: object mapping so that we can support objects as values
+        label: (options.find(({ value }) => value === val) as SelectFieldOption<V>)?.label ?? val,
+      })) ?? []
```
danad02, line 197: return value.map((val) => ({
      value: val,
      // we need to find instead of value: object mapping so that we can support objects as values
      label: (options.find(({ value }) => value === val) as SelectFieldOption<V>)?.label ?? val,
    }))

packages/ui/src/SelectField.tsx
```diff
@@ -154,6 +174,43 @@ const getMenuContainer = (menuPortalTarget: 'body' | 'self' | HTMLElement | null
...
+    )
```
danad02, line 198:
```suggestion
    return value.map((val) => ({
      value: val,
      // we need to find instead of value: object mapping so that we can support objects as values
      label: (options.find(({ value }) => value === val) as SelectFieldOption<V>)?.label ?? val,
    }))
```

packages/ui/src/SelectField.tsx
```diff
@@ -106,28 +106,45 @@ export type SelectFieldCoreDynamicProps<V> =
...
+// Since the user input is string, let's allow creatable only for string
+type SelectFieldCreatableProps<V> = V extends string
+  ? {
+      isCreatable?: boolean
+    }
+  : {
+      isCreatable?: false
+    }
+
+function isMultipleModeGuard<V>(value: V | V[], isMultiple: boolean): value is V[] {
```
danad02, line 140:
```suggestion
function isMultipleModeGuard<V>(value: V | V[] | null, isMultiple: boolean): value is V[] {
```

packages/ui/src/SelectField.tsx
jvorcak: Good catch @aigoncharov

packages/ui/src/SelectField.tsx
jvorcak: @danad02 but value can be `null` if you use multi mode combined with creatable

packages/ui/src/SelectField.tsx
```diff
@@ -184,6 +242,35 @@ export const SelectField = <V,>({
     }
   }, [menuPortalTarget])
 
+  const onChangeMultipleFn = React.useCallback(
+    (values: SelectFieldOption<V>[] | null) => {
+      ;(onChange as (newValue: V[] | null) => void)(values === null ? null : values.map(({ value }) => value))
+    },
+    [onChange],
+  )
```
danad02, line 250: Since empty `multi select` now returns an empty array when there's nothing selected, we might want to remove all these ugly `null`'s from other places:

```suggestion
  const onChangeMultipleFn = React.useCallback(
    (values: SelectFieldOption<V>[]) => {
      ;(onChange as (newValue: V[]) => void)(values.map(({ value }) => value))
    },
    [onChange],
  )
```

@jvorcak Can you also check other places too?

packages/ui/src/SelectField.tsx
```diff
@@ -106,28 +106,45 @@ export type SelectFieldCoreDynamicProps<V> =
...
+      value: V
+      onChange: (newValue: V) => void
     }
   | {
       isClearable: true
       isMulti: true
-      value: SelectFieldOption<V>[] | null
-      onChange: (newValue: SelectFieldOption<V>[] | null) => void
+      value: V[] | null
+      onChange: (newValue: V[] | null) => void
```
danad02, line 122: Since empty `multi select` now returns an empty array when there's nothing selected, we might want to remove all these ugly `null`'s from other places:

```suggestion
      value: V[]
      onChange: (newValue: V[]) => void
```

danad02 (commented):
@jvorcak I left few comments.

packages/ui/src/SelectField.tsx
danad02: @jvorcak It can't, I just tested it

danad02: @jvorcak Confirming that it's not an issue related to this PR. We have wrong types for `multi` from the beginning.

jvorcak: Screenshot 2021-03-04 at 9 40 50
Screenshot 2021-03-04 at 9 40 32

danad02: Well yes, you got it as a default value in the `useState`. If we change typings and you use `[]` as the default value, you won't get null there.

danad02: @jvorcak Sorry to drop this non-related issue on you during this great enhancement. I just think it's a good thing to fix it sooner rather than later.

jvorcak:
Thanks guys for a very great feedback. I've adjusted it, you can re-review @aigoncharov @danad02 @jankoritak 

- We now support `string` and `number` only as values
- Multiple mode can't have null, is `[]` instead (the same way as react-select`
- Direct access when transforming values
- Removed custom function for creatable mode

danad02 (approved):
Awesome! GJ :+1:

aigoncharov (approved):
LGTM!

PR #197, opened 2021-03-04
Feature/add titleless card

aigoncharov:


danad02 (approved):
Cool, just what we need :+1:

PR #198, opened 2021-03-04
Add missing "large" icon size implementation

aigoncharov:


danad02 (approved):
Wanted to do that several times. Always forgot :smile: 
GJ! :+1:

PR #210, opened 2021-03-09
update(Alert): fix issue with empty array as actions

fantkolja:
Fixes the following issue:
if you pass an empty array as the "actions" property of the `<Alert />` component, the component renders a "0" (aka `actions.length`) after the content.


jankoritak (approved):
Good catch!

aigoncharov (approved):

PR #214, opened 2021-03-12
Feature/interactive list

jvorcak:
- We want to store an array of strings in a Formik store
- We use <FieldArray /> https://formik.org/docs/api/fieldarray#validateonchange-boolean

However, our design works in a slightly different way, we want to trigger validation on submit. That is not how `<FieldArray />` works, we trigger the validation manually.


packages/ui/__stories__/InteractiveList.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { logger } from '@hazelcast/services'
+import { InteractiveListFormik } from '../src/InteractiveListFormik'
+import { List } from 'react-feather'
+
+export default {
+  title: 'Components/InteractiveList',
+  component: InteractiveList,
+  parameters: {
+    docs: {
+      inlineStories: false,
```
aigoncharov, line 14: nit: DO we need an iframe here?

packages/ui/src/InteractiveList.tsx
```diff
@@ -0,0 +1,121 @@
...
+          ariaLabel="Add Icon"
+          icon={Plus}
+          className={styles.addIcon}
+          size="normal"
+          onClick={async () => {
+            await addValueCallback()
+          }}
+        />
+      </div>
+      {CustomAddElement && <CustomAddElement setValue={setValue} />}
```
aigoncharov, line 97: Do you think we could limit it to a button for now? We could replicate `actions` in `ModalProps`, but make it a single optional action, not an array. for consistency. WDYT?

jvorcak: Sure no problem.

packages/ui/__stories__/InteractiveList.stories.tsx
jvorcak: Nope, I'll remove, thanks

packages/ui/src/InteractiveList.tsx
```diff
@@ -0,0 +1,121 @@
...
+        arrayHelpers.push(normalizedValue)
+        setValue('')
+      }
+    }
+  }, [value, normalizedValue])
+
+  return (
+    <>
+      <div className={styles.inputRow}>
+        <TextField
```
aigoncharov, line 67: @jvorcak I generally like the idea, but I am wondering if we could make the validation function totally custom. I would suggest to have 2 props:
- `validateCommitted` that validates the already added values in formik. It could be used to validate the array values.
- `validateUncommitted` that validates the current input value and receives the input value and the current array. It does not block the submission.

jvorcak: @aigoncharov I like the way where you want to go.

Basically `validateCommited` will not be needed, it'll be the default Formik validation (that will also work with `Yup` if present), that is how validation for `<FieldArray />` is meant - more about the problematic here -> https://formik.org/docs/api/fieldarray#fieldarray-validation-gotchas

The `validateUncommited` will be our custom function that will be used as a reaction to `<TextField/>` change, that is something Formik's FieldArray doesn't count on. WDYT?

aigoncharov: @jvorcak we'll still need to pass the default formik validation to the formik-bound field, right? So `validateCommited` is meant for `InterqctiveListFormik`.

packages/ui/src/InteractiveList.tsx
jvorcak: So I've had a look and passing a component is probably the best solution? The point is we pass this `setValue` pointer that can change the internal state of our component. I'm not sure how this would be doable with approach like we see in ModalProps.

packages/ui/src/InteractiveList.tsx
jvorcak: @aigoncharov btw, what is the advantage of having a separate validation for the current input? Why can't we just use the `validate` that is passed to a Formik's `useField`?

aigoncharov: @jvorcak good idea! Let's do exactly that!

packages/ui/src/InteractiveList.tsx
aigoncharov: What if we exposed `setValue` via `useImperativeHanlde` and let any arbitrary JSX here? Then we could have more freedom in terms of how we can control the state of InteractiveList and at the same time we could use the `setValue` from the exposed handle in the JSX that we wrap. WDYT?

packages/ui/src/InteractiveListFormik.tsx
```diff
@@ -29,15 +36,18 @@ export const InteractiveListFormik = <V extends object>({ name, validate, ...pro
       name={name}
       render={(arrayHelpers) => {
         return (
-          <InteractiveList<V>
+          <InteractiveList
+            ref={inputRef}
```
aigoncharov, line 40: I would suggest to use a custom prop, say, `inputStateControlsRef` instead of a forwarded ref as forwarded refs are usually point to the dom elements. WDYT?

jvorcak: You're totally right!

packages/ui/src/InteractiveList.tsx
```diff
@@ -0,0 +1,169 @@
...
+      setValue(value)
+    },
+  }))
+
+  const addValueCallback = useCallback(async () => {
+    if (normalizedValue.length === 0) {
+      onError('You need to provide a non empty value')
+    } else if (value.includes(normalizedValue)) {
+      onError('You need to provide a unique value')
+    } else {
```
aigoncharov, line 100: @jvorcak should this be a part of `validate`?

aigoncharov: Since we can display the erro in a single place only - under the top input, should we also change our stories to always return a string or undefined from the validator? In other words, what sense does it make to returned structured errors for every item in the array when we do not diaply them? Moreover, shouldn't;t we rather validate our input value before we add it to the list, so there are no errors in the list ever?

jvorcak: @aigoncharov this is the problematic part. We show errors also per item, this is simply a requirement from the Formik side in case we use FieldArray with it's helpers. The reason for this is that Formik has built-in `yup` validation mechanism, and that can provide your FieldArray component with an array of errors, the component should support it. There is no other way around it except for forbidding Yup completely somehow. 

The problematic part is that there is no way (or I haven't found any) how to run a `Yup` validation for this specific field on demand by providing a value (from our custom field) and getting a result (that could prevent from entering a value).

So this solution is a fallback, if you provide validate, we will guard you on entering the value, and support `yup` validation as a last check before submitting the value -> so basically the UX is not ideal if Yup validation contains more checks than custom validate. This is what I tried to point out initially, hopefully it's more clear now.

aigoncharov: @jvorcak do you use schema-based validation in Cloud?

jvorcak: And with Yup, you'll need to make sure to provide it explicitelly via validate to get a better UX. Not exactly ideal, but the best I could think of with Formik

aigoncharov: @jvorcak maybe, I am overthinking this. Let's stick to the current implementation. One thing though. It does not re-validate the input value on change for me in storybook. Could you check that?

jvorcak: I don't think you are, I'm not 100% happy with it either, but I've spent enough time on investigating how it could be done better, and this is the best I could think of.

aigoncharov (approved):

aigoncharov:
@jvorcak one last idea:
1. Use `useFormik` to get `validateForm` and `values` in the `InteractiveListFormik`.
2. Add `onChange` listener to `InteractiveList`
3. Run `validateForm` with the updated values with added input value
4. Scan for errors in the array. If any found, return the first one as the input error.

jvorcak:
@aigoncharov that's a life safer, luckily validateForm can receive an optional values (unlike `validateFormField`). Can you please check now?

I've also made a validation on change, but only after an unsuccessful submit. WDYT?

jvorcak:
@aigoncharov I'd still keep the ability of Formik to provide an array of errors (e.g. validation fails on a server)

packages/ui/src/InteractiveListFormik.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { validateForm } = useFormikContext<V>()
+  const [field, meta, { setTouched, setError }] = useField<string[]>({
+    name,
+    validate,
+  })
+
+  const onError = useCallback(
+    (value: string | string[]) => {
+      setTouched(true, false)
+      // @ts-expect-error Formik expects the string[], while documentation says
```
aigoncharov, line 40: nit: could we use simple `as`?

packages/ui/src/InteractiveList.tsx
```diff
@@ -0,0 +1,207 @@
...
+    if (touched) {
+      void getValidationError().then((validateError) => {
+        if (validateError) {
+          onError(validateError)
+        }
+      })
+    }
+  }, [touched, inputValue])
+
+  const getValidationError = async () => {
```
aigoncharov, line 110: @jvorcak the only thing that is bothering me just a bit is that now this component is Fromik-bound. What if we lift validation and error handling to `InteractiveListFormik`? Then this component will accept the final error as `error` prop and to trigger the validation chain in parent `InteractiveListFormik` we could expose `onChange`. WDYT?

packages/ui/__stories__/InteractiveListFormik.stories.tsx
```diff
@@ -0,0 +1,118 @@
+import React, { useRef } from 'react'
+import { Button } from '../src'
+import { InteractiveListInputRef } from '../src/InteractiveList'
+import { Form, Formik } from 'formik'
+import { logger } from '@hazelcast/services'
+import { InteractiveListFormik } from '../src/InteractiveListFormik'
+import { List } from 'react-feather'
+import * as Yup from 'yup'
```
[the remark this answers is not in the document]
jvorcak: This is just storybook, I don't think it's relevant here.

packages/ui/src/InteractiveList.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  useEffect(() => {
+    if (touched) {
+      void getValidationError().then((validateError) => {
+        if (validateError) {
+          onError(validateError)
+        }
+      })
+    }
+  }, [touched, inputValue])
```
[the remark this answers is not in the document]
jvorcak: @fantkolja I think this would be valid point in `useCallback`. In `useEffect` I think it's perfectly normal to define when a hook should be run like this, and those pointers doesn't need to be used inside a function directly. Our ESlint rules says so as well, they will throw an error like you mention just for `useCallback`.

Regarding `getValidationError`, it will get the correct value. Like in [React's documentation ](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects) they don't include `ChatAPI` in a dependencies, that is the case for `useCallback`

packages/ui/src/InteractiveListFormik.tsx
jvorcak: @aigoncharov well, the real value is `string`, but Formik expects `string[]`, so I think `as` is not usable here?

packages/ui/src/InteractiveList.tsx
jvorcak: The reason I went this way is that `addValueCallback` - function that adds an actual element to a list uses `arrayHelpers` from Formik -> thus needs to be in a lower component. In this function, I'd like to call validation somehow to make sure the value I add is correct. I could use it as a prop from outer Formik component, but that would make it Formik-bound anyway -> just a bit more complicated.  Or do you have some other idea?

I think if a component is dependent on `arrayHelpers`, it's formik-bound either way.

aigoncharov: Makes sense. Another option wild be to provide an onValueAdd callback to the parent formik-bound component. Which could run validation and sdd the value if needed. This way the list component could be used with any form library in the future. WDYT?

jvorcak: @aigoncharov via ImperativeHandle? Isn't that overusing is bit? However, I can do that if you want.

jvorcak: @aigoncharov you can have a look

aigoncharov: @jvorcak I was about to propose something similar to what you have implemented. Great job!

aigoncharov (approved):

aigoncharov (approved):

packages/ui/src/InteractiveList.tsx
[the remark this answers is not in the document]
jvorcak: @fantkolja I've added those dependencies, it shouldn't hurt anything, the pointers will never change in our case, and we do have an eslint warning for that.

packages/ui/src/InteractiveList.tsx
```diff
@@ -0,0 +1,135 @@
...
+          icon={Plus}
+          className={styles.addIcon}
+          size="normal"
+          onClick={onAddItem}
+        />
+      </div>
+      {children}
+      <ul className={styles.list}>
+        {value.map((str, idx) => (
+          <li key={idx}>
```
danad02, line 121: Since values are unique I suggest to use them as keys:
```suggestion
        {value.map((str, idx) => (
          <li key={str}>
```

packages/ui/src/InteractiveListFormik.tsx
danad02: It is usable and works as expected:
```
setError(value as string[])
```

danad02 (approved):
A bit late to the party :balloon: 
Anyway I like the various changes that you added during review. I did some testing and it looks good :+1:

packages/ui/src/InteractiveList.tsx
jvorcak: @danad02 I've accepted as it makes it more bulletproof, thanks. However this was recommended way by FieldArray and it worked with non-duplicate changes, so FieldFormik took care of this.

packages/ui/src/InteractiveList.tsx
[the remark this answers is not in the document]
fantkolja: I believe the ChatAPI in the example above is taken from the outer scope and btw they have left a note about including all the deps.
image

I think everything will work if you update the `useEffect()` a bit to prevent looping
`  useEffect(() => {
    if (inputTouched) {
      void getValidationError().then((validationError) => {
        // prevent looping
        if (validationError && validationError !== meta.error) {
          onError(validationError)
        }
      })
    }
  }, [inputTouched, inputValue, getValidationError, onError])`
  
And btw with this code snippet the error message is not removed after submit

packages/ui/src/InteractiveListFormik.tsx
jvorcak: @danad02 I'm not very comfortable doing `as string[]` when I know the value is `string`

packages/ui/src/InteractiveList.tsx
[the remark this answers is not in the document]
jvorcak: @fantkolja this impl, gives me an infinite loop, even if I wrap getValidationError with `useCallback`

fantkolja (approved):

PR #227, opened 2021-03-25
update(datepicker): remove react-datepicker from peer deps

fantkolja:
In my opinion, the additional peer dependency makes the lib less flexible and will make future upgrades more difficult.
This is a suggestion on how to omit it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,16 @@ However, Next.js still [forbids](https://github.com/vercel/next.js/blob/master/e
 So please make sure to import the following global styles in your projects if you need them.
 
 ```typescript
-import 'react-datepicker/dist/react-datepicker.css'
+import '@hazelcast/ui/styles/datepicker.scss'
 ```
+
+or
+
+```typescript
+import '@hazelcast/ui/styles/datepicker.module.scss'
+```
+
+if you need a CSS modules version
 
 ## SSR
 
--- a/packages/ui/.storybook/preview.scss
+++ b/packages/ui/.storybook/preview.scss
@@ -1,5 +1,6 @@
 @use '../styles/fonts';
 @use '../styles/constants' as c;
+@use '../styles/datepicker/datepicker.module';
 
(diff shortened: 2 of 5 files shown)


aigoncharov:
@fantkolja have you tried it? I experimented with something like it. Since npm tries to keep the folder structure flat,  `react- datepicker` is usually installed not in the child folder of `hazelcast/ui`, but in the neighbor folder.
So in your project you will most probably have something like this:
- yourProject 
  - node_module
    - @hazelcast/ui
    - react-datepicker 
    
As you can see, react-datepicker and hazelcast/ui are on the same level. To my best knowledge, SCSS wasn't able to resolve react-datepicker stylesheet in this folder layout. Does it work for you in the cloud?

fantkolja:
> @fantkolja have you tried it? I experimented with something like it. Since npm tries to keep the folder structure flat, `react- datepicker` is usually installed not in the child folder of `hazelcast/ui`, but in the neighbor folder.
> So in your project you will most probably have something like this:
> 
> * yourProject
>   
>   * node_module
>     
>     * @hazelcast/ui
>     * react-datepicker
> 
> As you can see, react-datepicker and hazelcast/ui are on the same level. To my best knowledge, SCSS wasn't able to resolve react-datepicker stylesheet in this folder layout. Does it work for you in the cloud?

But since it uses webpack modules resolving mechanism ("borrowed") from NodeJS it looks up the `node_modules` going upwards first trying to resolve the package in those folders. so in your scheme, it will first find `yourProject/node_modules` and then resolve `react-datepicker` in that folder.

Tried with our app (by importing `import '@hazelcast/ui/styles/datepicker.scss';` and it worked, styles were applied.

aigoncharov:
@danad02 @jankoritak @utkukaratas  could you try this fix before merging it in MC first?

jankoritak:
Seems to be working fine on MC-side. 👍

jvorcak (approved):

PR #231, opened 2021-04-07
update(Link): add possibility to pass ref to an Anchor Link component

fantkolja:
Using with `next/link` it's required to pass the ref to the underlying `<a>` component.
Check [here](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-function-component) for more info.


packages/ui/src/Link.tsx
```diff
@@ -80,43 +80,54 @@ export type LinkProps = IconProps & {
...
-  ariaLabel,
-  iconClassName,
-  href,
-  rel = 'noopener',
-  target = '_self',
-  onClick,
-  className,
-  children,
-}) => {
+export const Link = React.forwardRef<HTMLAnchorElement, LinkProps>(function Link(
```
aigoncharov, line 83:
```suggestion
export const Link = React.forwardRef<HTMLElement, LinkProps>(function Link(
```
Then we could pass refs to the button as well. WDYT?

fantkolja: @aigoncharov I agree, that would enhance the component, but I can't define proper typing this way, I keep getting type errors here
image

The only way I could think of to properly type the component is to actually make a separate component LinkButton...
Maybe you could help me...

aigoncharov: @fantkolja pushed a fix

aigoncharov (approved):

Issue #234, opened 2021-04-14
Disappearing validation error on submit


fantkolja:
If the user clears the input n the `<NumberFieldFormik />` and then submits the form the `touched` property will be set to `false` which will take away the validation error if present

fantkolja:
Caused by this Fromik's issue https://github.com/formium/formik/issues/2332

PR #235, opened 2021-04-14
Fix/hzfs 234 disappearing validation error

fantkolja:
Closes https://github.com/hazelcast/frontend-shared/issues/234


aigoncharov (approved):
LGTM!
BTW, could we have a test for this bug?

fantkolja:
> LGTM!
> BTW, could we have a test for this bug?

@aigoncharov Agree. Test added.
Thanks again for your help!

PR #236, opened 2021-04-19
Enhance Card

aigoncharov:


fantkolja (approved):
LGTM!
nit: would be good to add stories for new props

PR #239, opened 2021-04-22
Add Code and CodeEditor components based on Codemirror 6

### LINKED ISSUE #71: [Core UI] Code
Add support for the `Code` element

utkukaratas:
closes #71


utkukaratas:
guys any help here appreciated @aigoncharov @fantkolja   

all package-locks generated via npm 7. code is freshly merged. i'm at a loss for the reason why the pr builder fails.

```18:42:05 > @hazelcast/ui@0.1.0-alpha.0 compile /home/jenkins/jenkins_slave/workspace/frontend-shared-pr/packages/ui
18:42:05 > shx rm -f *.tsbuildinfo && shx rm -rf lib && shx mkdir lib && shx cp -r src/* lib && shx rm -rf lib/*.{tsx,ts} && shx rm -rf lib/**/*.{tsx,ts} && tsc -p tsconfig.build.json
18:42:05 
18:42:05 src/Dialog.tsx(53,9): error TS2322: Type '{ onClick: MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>; children: string; disabled?: true | undefined; disabledTooltip?: string | undefined; autoFocus: boolean; kind: "primary" | "danger"; }[]' is not assignable to type 'ModalActionProps[]'.
18:42:05   Type '{ onClick: React.MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>; children: string; disabled?: true | undefined; disabledTooltip?: string | undefined; autoFocus: boolean; kind: "primary" | "danger"; }' is not assignable to type 'ModalActionProps'.
18:42:05     Type '{ onClick: React.MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>; children: string; disabled?: true | undefined; disabledTooltip?: string | undefined; autoFocus: boolean; kind: "primary" | "danger"; }' is not assignable to type '{ kind?: ButtonKind | undefined; children: string; capitalize?: boolean | undefined; bodyClassName?: string | undefined; outlineClassName?: string | undefined; outline?: ButtonOutlineType | undefined; } & ... 5 more ... & Pick<...>'.
18:42:05       Type '{ onClick: React.MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>; children: string; disabled?: true | undefined; disabledTooltip?: string | undefined; autoFocus: boolean; kind: "primary" | "danger"; }' is not assignable to type 'ButtonDisabledProps'.
18:42:05         Types of property 'disabledTooltip' are incompatible.
18:42:05           Type 'string | undefined' is not assignable to type 'string'.
18:42:05             Type 'undefined' is not assignable to type 'string'.
```

packages/icons/package-lock.json
```diff
@@ -1,14536 +1,8 @@
 {
   "name": "@hazelcast/icons",
   "version": "0.1.0-alpha.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
aigoncharov, line 4: Could you use npm 7 to keep lockfileVersion to 2?

packages/ui/__tests__/Code.test.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import { Code } from '../src/Code'
+
+describe('Code', () => {
+  it('Renders the default Code correctly', () => {
+    const wrapper = mount(<Code value="Hello World" />)
+
+    // following does not work because of how CodeMirror creates itself
```
aigoncharov, line 10: Why does not it work? Could you be a bit more specific?

packages/ui/package.json
```diff
@@ -43,6 +43,10 @@
     "url": "https://github.com/hazelcast/frontend-shared/issues"
```
aigoncharov, line 43: Why the associated package-lock was removed?

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  We are not using the old CM5; we're basing our component on CM6 which is the next gen. version of the hugely
+  popular CM5. CM6 is still years old and solid despite being a young project.
+
+  See https://codemirror.net/6/ for more.
+
+  ## About language support
+
+  In CM6 all of the languages are defined in small, self defined packages.
+  ie. import { javascript } from '@codemirror/lang-javascript'
```
aigoncharov, line 16: I would suggest we use the modern packages as well to propagate the best practices. Cloud also cares about the bundle size as well.

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+  So using a language is one import and one property assignment. See the `CodeEditor.stories.tsx`.
+
+  for more see https://github.com/codemirror/legacy-modes
+
+  ## About styling/themes
+
+  CM6 is not that rich with predefined themes yet so we are not providing anything other than the
+  default theme (though it's easy provide custom ones if/when we need it.)
+
+  ## Customizing the editor
```
aigoncharov, line 32: Should we customize the default theme with our colors and our fonts?

packages/ui/src/CodeEditor.tsx
utkukaratas: hmm that makes it harder to use (ok not hard but unpleasant)... you have to add the language dependencies one by one to your project then. do we want that?

packages/ui/src/CodeEditor.tsx
utkukaratas: another case of "do we want that?". in MC (sql-browser) we're using one of the default themes (solarized). i could backport that if you think we should.

packages/ui/src/CodeEditor.tsx
aigoncharov: Absolutely! This will enable tree-shaking on Cloud

packages/ui/src/CodeEditor.tsx
aigoncharov: @utkukaratas I would like the code editor to use our new fonts and colors, not the old ones. Just for it to look consistent with the rest of the UI. If you are in doubt, I would suggest consulting with @pawelklasa. However, we might want to prototype something first ourselves not to be blocked by him.

utkukaratas: luckily the default theme is inline with our new MC colors with the white/blue/grays. let me show you and pawel things live when redesign work is done this week and we could backport any theme if we don't like it. resolving this.

packages/ui/__tests__/Code.test.tsx
utkukaratas: while documenting it, i found a neat trick to resolve it :)) resolved.

packages/ui/src/CodeEditor.tsx
utkukaratas: hmm i see no way how tree-shaking would be ineffective with the current code (it's ONE package yes; but it's MANY language modules, you import the ones you need - so bundle size would not be that much different than our current code.)

i'm still reluctant to revert using the new packages (i did in the beginning; did not like it at all) imho we should roll with this.

packages/ui/src/CodeEditor.tsx
aigoncharov: Are they *our* gray, blue and green though? What about the font?

packages/ui/src/CodeEditor.tsx
fantkolja: Do I understand it correctly, on the client we can do both: import legacy as well as modern language packages, (i.e. '@codemirror/legacy-modes/mode/javascript' or '@codemirror/lang-javascript')?
If so then the client will care about how big the language import is, since here we only use it in stories so the language won't be a part of the library bundle. The only drawback is one more dependency but it's not that bad.
If so then please update the comments, because when I read them I'm under impression that only the legacy languages are allowed...

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+}
+
+const DEFAULT_OPTIONS: CodeOptions = {
+  language: null,
+  lineWrapping: false,
+  lineNumbers: true,
+}
+
+export const CodeEditor: FC<CodeEditorProps> = ({ value, className, options = {}, onChange, customExtensions, innerRef }) => {
+  const parentRef = useRef<HTMLDivElement>(null)
```
fantkolja, line 95:
```suggestion
  const parentRef = useRef<HTMLDivElement | null>(null)
```

like the line below?

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+        parent: parentRef.current,
+      })
+
+      return () => {
+        // destroy cm. (mind you; this component will be recreated whenever options is modified.)
+        cm.current?.destroy()
+        cm.current = null
+      }
+    },
+    // value and onChange should not be a dependencies, just disable the warnings:
```
fantkolja, line 168: But they are dependencies. I believe it's a bad practice just to suppress such warnings, which might result in hard to catch bugs

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+export const CodeEditor: FC<CodeEditorProps> = ({ value, className, options = {}, onChange, customExtensions, innerRef }) => {
+  const parentRef = useRef<HTMLDivElement>(null)
+  const cm = useRef<EditorView | null>(null)
+
+  const optionsMemoized = useDeepCompareMemo(() => options, [options])
+
+  // use this handle to access the cm element (which is a `view`)
+  useImperativeHandle(innerRef, () => ({
+    view: () => cm.current!,
```
fantkolja, line 102: I'm not sure about the non-null assertion operator (!), we are setting the current value only in the next `useEffect()` , manually (!!!) and only after the `parentRef` has its `.current`, we may have here `null`

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+  }))
+
+  useEffect(
+    function initCodeMirror() {
```
fantkolja, line 106: nit: is there a reason for a non-arrow function declaration here? This `function` hoists...

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,178 @@
...
+      cm.current = new EditorView({
+        state: EditorState.create({
+          doc: value || '',
+          extensions,
+        }),
+        parent: parentRef.current,
+      })
+
+      return () => {
+        // destroy cm. (mind you; this component will be recreated whenever options is modified.)
```
fantkolja, line 163:
```suggestion
        // destroy cm. (mind you; this component will be recreated whenever options are modified.)
```

fantkolja (commented):
Nicely done!
Few comments

packages/ui/src/CodeEditor.tsx
utkukaratas: nice catch. resolving...

packages/ui/src/CodeEditor.tsx
utkukaratas: just for cleaner stacktraces.

packages/ui/src/CodeEditor.tsx
utkukaratas: it was there to make typescript compiler happy; now gone; i found a better way on a second look. resolving...

packages/ui/src/CodeEditor.tsx
utkukaratas: that's the point - i don't want them to be dependencies. code would be more convoluted without that.

packages/ui/src/CodeEditor.tsx
utkukaratas: another nice catch! resolving.

packages/ui/src/CodeEditor.tsx
utkukaratas: @fantkolja @aigoncharov i think my documentation confused you guys; i've touched it a bit - not mentioning "the new language modes" at all (we have no reason to use them; legacy pack gives us all we need and more). 

@fantkolja yes you're right. you can use both legacy and new language modes*. legacy language pack is already a dependency of core-ui so you WON'T NEED to add it to your dependencies manually and your bundle will contain ONLY the languages you decide to use  - that's the beauty of it. 

* you are allowed to use new language modes but then you will need to add them to your dependencies one by one. and then there're not that many of them implemented at the moment. so why bother? (you might be triggered by the word "legacy"; don't worry about that they're not going away soon.)

packages/ui/src/CodeEditor.tsx
fantkolja: But in this case, if the parent component changes `value` or `onChange` props after initialization the updates will be ignored...

utkukaratas: сука блять

utkukaratas: :))) now that's a valid concern @fantkolja thx, will handle that.

packages/ui/src/CodeEditor.tsx
aigoncharov: nit: I would suggest to use an arrow function here for consistency.

packages/ui/src/CodeEditor.tsx
utkukaratas: @fantkolja eslint-disable-next-line is still there for .value but the case is handled and documented. resolving...

utkukaratas:
@aigoncharov @fantkolja hi folks, all of your comments must be handled, another round of reviews pls.

package-lock.json
```diff
@@ -6,13 +6,17 @@
     "": {
       "name": "root",
       "dependencies": {
+        "@codemirror/basic-setup": "0.18.0",
+        "@codemirror/legacy-modes": "0.18.0",
```
aigoncharov, line 10: Should this one be a devDependency?

packages/icons/package-lock.json
```diff
@@ -14,6 +14,7 @@
         "@storybook/theming": "^6.1.21",
         "@types/react": "16",
         "@types/react-dom": "16",
+        "loki": "^0.28.1",
```
aigoncharov, line 17: Why do we need to touch this package at all? Have we added any new icons?

packages/ui/package.json
```diff
@@ -43,6 +43,10 @@
...
   },
   "dependencies": {
+    "@codemirror/view": "0.18.8",
+    "@codemirror/basic-setup": "0.18.0",
+    "@codemirror/legacy-modes": "0.18.0",
```
aigoncharov, line 48: Should this one be a dev dependency?

packages/ui/src/CodeEditor.module.scss
```diff
@@ -0,0 +1,17 @@
+@use '../styles/constants' as c;
+@use '../styles/helpers' as h;
+
+.container {
```
aigoncharov, line 4: @utkukaratas could you get @pawelklasa to sign off on the colors and fonts?

packages/ui/src/CodeEditor.tsx
```diff
@@ -0,0 +1,194 @@
...
+      })
+
+      return () => {
+        // destroy cm. (mind you; this component will be recreated whenever options are modified.)
+        cm.current?.destroy()
+        cm.current = null
+      }
+    },
+    // value should not be a dependency (it is handled below), just disable the warnings:
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
aigoncharov, line 174: Do we still need to disable the linter check on this line?

utkukaratas: @aigoncharov yes, for .value.

packages/ui/package.json
utkukaratas: @aigoncharov no, why would be want that? it's just convenient to have it there instead of your app code. (see the unresolved comments at the top.)

packages/icons/package-lock.json
utkukaratas: @aigoncharov  i have no idea why that changed at all. half of the time i have fought with npm here :/ there were many conflicts and what not so far; i stumbled upon a working version and stuck with it.

packages/ui/src/CodeEditor.tsx
aigoncharov: @utkukaratas do you mean that the legacy languages package is three-shakable?

packages/icons/package-lock.json
aigoncharov: @utkukaratas do you think we could revert this change as it introduces an absolutely unnecessary dependency to a package we do not even touch in this PR?

packages/ui/src/CodeEditor.tsx
utkukaratas: @aigoncharov yes, those languages are defined in small self contained modules each.

packages/icons/package-lock.json
utkukaratas: will try for sure.. and i was afraid you're going to say that 😬

aigoncharov: Sorry :(

packages/ui/src/CodeEditor.module.scss
utkukaratas: pawel gives the green. resolving.

packages/icons/package-lock.json
utkukaratas: @aigoncharov done. luckily it was not big deal; resolving.

aigoncharov (approved):
Great job! Feel free to merge it once @pawelklasa sings off on the fonts and colors!

fantkolja (approved):

PR #261, opened 2021-05-20
Feat: add smallMedium icon size

danad02:
Adds `smallMedium` icon size.
Rename `normal` icon size to `medium`.


danad02:
@aigoncharov @utkukaratas @fantkolja Since we're making this change - should we also change naming in other components? We use `normal` almost everywhere and only Icon now has `medium` instead.

aigoncharov (approved):

PR #264, opened 2021-05-24
Feat: update Pagination according to design

danad02:
Update Pagination according to latest design.

`TextField` and `SelectField`:

- Add `showAriaLabel` prop of type `boolean`. When `true` we don't show `<Label />` component, `aria-label` on the underlying `input` is used instead. Defaults to `false`.

`NumberField`:

- Add `showIconButtons` prop of type `boolean`. When `false` increment and decrement buttons are hidden. Defaults to `true`,


aigoncharov (approved):
LGTM!

PR #282, opened 2021-06-01
Feat/code editor issues

### LINKED ISSUE #259: CodeEditor: several issues
- changing border-radius is problematic (due to many layers of elements Codemirror use internally. anyway the lib should handle this.)
- missing: formik v2 wrapper.
- missing: .error prop.
- we need .rows prop like a textarea. hugely convenient.

utkukaratas:
closes #259


Sodik (approved):

packages/ui/src/CodeEditor.tsx
```diff
@@ -147,13 +185,26 @@ export const CodeEditor: FC<CodeEditorProps> = ({ value, className, options = {}
         ]),
       ]
 
+      const fixedHeightExtension = () => {
+        if (!opts.rows) return []
+
+        const LINE_HEIGHT = 22
+        const h = LINE_HEIGHT * opts.rows + 4
```
danad02, line 192: nit: what does the `+ 4` stand for? Some free space at top and bottom?

utkukaratas: yep, and let me make that a constant.

PR #286, opened 2021-06-03
feat: Table global search

Sodik:


fantkolja (approved):
LGTM!

jgardiner68:
Is this searching the rendered values or the data as supplied by the backend?  We have an issue in MC where the search works on the BE data rather than the rendered values which is confusing for the user.  https://hazelcast.atlassian.net/browse/MC-785

Sodik:
> Is this searching the rendered values or the data as supplied by the backend? We have an issue in MC where the search works on the BE data rather than the rendered values which is confusing for the user. https://hazelcast.atlassian.net/browse/MC-785

This is only for rendered values. I'm going to use it on Maps page

PR #296, opened 2021-06-08
feat: Table header can be hidden

Sodik:
feat: Table header can be hidden
fix: Hide pagination when table does not have data


utkukaratas (approved):
yay!

danad02 (approved):
LGTM

PR #315, opened 2021-06-17
fix: Overlay header text overlapping close button

Sodik:


packages/ui/src/Overlay.module.scss
```diff
@@ -55,10 +55,21 @@
...
     .header {
+      $iconSize: c.$grid * 9;
+
       display: flex;
       justify-content: center;
       align-items: center;
       margin-bottom: c.$grid * 15;
+      padding: 0 c.$grid * 21; // close button width
+
+      @media only screen and (max-width: 640px) {
```
fantkolja, line 66: nit: Should we add the size to a constant, wdyt?

fantkolja (approved):

packages/ui/src/Overlay.module.scss
Sodik: I'm not sure, his is not fixed width. I just get it from devtools layout with extra space to compensate fonts appearance in different browsers

utkukaratas (approved):

PR #320, opened 2021-06-18
move @codemirror/legacy-modes dep to devDependencies.

utkukaratas:
initially i thought declaring this module in hive's dependencies would be just alright - letting one use legacy-modes package without adding it to the project's dependencies. but as it turns out you'll need to explicitly add @codemirror/legacy-modes to your project; otherwise you get a warning. 

so let's move it to devDependencies (stories use it).

ps. also fixes the bogus default .rows prop.


Sodik (approved):

danad02 (approved):

utkukaratas:
verify

PR #323, opened 2021-06-21
feat: Flexible Button variants

Sodik:
Replace property kind by combination of variant and color properties. Property kind will be removed later.


packages/ui/__stories__/Button.stories.tsx
```diff
@@ -156,3 +156,53 @@ export const LinkSemantics = () => {
     </>
   )
 }
+
+export const AllVariants = () => {
```
danad02, line 160: Can we just update the `Template` with all combinations of `variant` and `color`? Other stories are bound to the template and will update accordingly.

packages/ui/__stories__/Button.stories.tsx
```diff
@@ -156,3 +156,53 @@ export const LinkSemantics = () => {
...
+
+        <Button {...props} variant="outlined" color="primary" />
+        <Button {...props} iconLeft={Info} iconLeftAriaLabel={'Additional information'} variant="outlined" color="primary" />
+        <Button {...props} variant="outlined" color="secondary" />
+        <Button {...props} iconLeft={Info} iconLeftAriaLabel={'Additional information'} variant="outlined" color="secondary" />
+
+        <Button {...props} variant="text" color="primary" />
+        <Button {...props} iconLeft={Info} iconLeftAriaLabel={'Additional information'} variant="text" color="primary" />
+        <Button {...props} variant="text" color="secondary" />
+        <Button {...props} iconLeft={Info} iconLeftAriaLabel={'Additional information'} variant="text" color="secondary" />
```
danad02, line 181: Is there a difference between `variant="text" color="primary"` and `variant="text" color="secondary"`?

danad02 (changes requested):
I like this change of props! Leaving two outstanding comments.

packages/ui/__stories__/Button.stories.tsx
Sodik: Text inherits color so there is no deference. Maybe we will need to change it

danad02 (approved):
LGTM :+1:

PR #334, opened 2021-06-25
feat: formik field wrappers onChange callback

Sodik:


danad02 (changes requested):
Nice, this is something we will definitely use in some scenarios! Could you also add the `onChange` prop to `NumberFieldFormik` and `MultiSelectFieldFormik`?

danad02 (approved):
LGTM!

fantkolja (approved):
Well done!!!

PR #347, opened 2021-07-13
Refactor: change Card's margin bearers, update Card stories

danad02:


packages/ui/src/Card.module.scss
```diff
@@ -11,25 +11,21 @@
...
       margin-right: c.$grid * 2.5;
     }
 
     .title {
       margin: 0;
-
-      &.space {
-        margin-right: c.$grid * 2.5;
-      }
+      margin-right: c.$grid * 2.5;
```
Sodik, line 21: I wouldn't change this. Extra 10px on a small screen is priceless. What do you thinks @utkukaratas @fantkolja  @allfayn?

Sodik (approved):

packages/ui/src/Card.module.scss
danad02: @Sodik You're right, I wanted to revert this. Here it is: https://github.com/hazelcast/hive/pull/347/commits/b107f6a6cee6705a5dbe0e950e82aecc7c7c7ba0

PR #351, opened 2021-07-19
update(ui): handle dynamic min on Slider

### LINKED ISSUE #345: Slider issues
1. Marks and value labels should be ReactNode not string

2. The value is not recalculated when the min value is changed and the value is below the min value
image

3. Mark has a broken layout
image

fantkolja:
This is only first part, since I need to deploy new version of `@hazelcast/helpers` in order to use it in Slider


packages/ui/src/Slider.tsx
```diff
@@ -57,6 +58,29 @@ function isSingleValueGuard(value: SliderValue, onChange?: SingleValueChangeFn |
   return !Array.isArray(value)
 }
 
+function triggerNativeInputChange(value: string, inputEl: HTMLInputElement) {
```
Sodik, line 61: I would move it to helpers and make it a generic function to fire an event on any element

Sodik (changes requested):

allfayn (approved):

packages/ui/src/Slider.tsx
fantkolja: Moved to helpers. But now we need to deploy new helpers version in order to use it in ui. So let's merge this one and then I will create another PR

Sodik (approved):

PR #356, opened 2021-07-27
feat: CheckableSelectField - Custom MultiSelectField

Sodik:


packages/ui/__stories__/CheckableSelectField.stories.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+
+export const OpenWithInputValue = Template.bind({})
+OpenWithInputValue.args = {
+  ...Empty.args,
+  ...Open.args,
+  inputValue: 'obi',
+}
+
+export const OpenWithInputValueAndNoResults = Template.bind({})
```
fantkolja, line 104: When there is an input value, some options are filtered out and I click 'Select All' also the hidden options are selected. This doesn't feel right
image

packages/ui/src/Select/CheckableSelectField.module.scss
```diff
@@ -0,0 +1,36 @@
+@use './styles/constants' as c;
```
fantkolja, line 1: Seems like the checkbox is higher than the text
image

fantkolja (changes requested):
Well done!
Few comments...

utkukaratas (approved):
overlaps a lot with MultiSelectField :/

Sodik:
I think we should review selects and perhaps remove react-select or refactor this component to make it more compatible with react-select way @pawelklasa

packages/ui/__stories__/CheckableSelectField.stories.tsx
```diff
@@ -0,0 +1,137 @@
+import React, { useState } from 'react'
```
fantkolja, line 1: nit: Probably we should wrap long options, wdut?
image

fantkolja (approved):

PR #363, opened 2021-08-17
feat: Render Datepicker into body

Sodik:


utkukaratas (approved):

packages/ui/src/Calendar/Calendar.tsx
```diff
@@ -1,17 +1,27 @@
...
 const DATE_FORMAT_WITH_TIME = 'yyyy-MM-dd hh:mm a'
 
-const CalendarPopperContainer: FC = ({ children }) => <div data-test="date-picker-popper-container">{children}</div>
+const CalendarPopperContainer: FC = ({ children }) =>
+  children
+    ? createPortal(
+        <div data-test="date-picker-popper-container" className="hz-date-picker-popper-container">
+          {children}
+        </div>,
+        canUseDOM ? document.body : ((null as unknown) as HTMLElement),
```
fantkolja, line 22: seems too hacky...
how about we move this condition to the `children` condition?

`children && canUseDOM ? createPortal() : null` ?

fantkolja (changes requested):

packages/ui/src/Calendar/Calendar.tsx
Sodik: done

fantkolja (approved):

PR #373, opened 2021-09-13
update(ui): use small button size as default and remove property

### LINKED ISSUE #372: Use only small Button
Remove property 'size' for Button component and use small size by default

allfayn:
Closed #372


utkukaratas (approved):

fantkolja (approved):

packages/ui/src/Button.module.scss
```diff
@@ -20,17 +20,17 @@
...
-  border-radius: c.$borderRadius;
-  height: c.$inputHeight;
+  border-radius: c.$borderRadiusButtonSmall;
+  height: c.$inputHeightSmall;
   max-width: c.$buttonMaxWidth;
   // for <a> element
   text-decoration: none;
 
-  @include c.typographyButton;
+  @include c.typographyButton('small');
```
Sodik, line 29: Why do we have to pass a size if there is only one size?

PR #402, opened 2021-11-02
feature(ui): NumberField with a separate icons view

allfayn:
New type for a view plus/minus buttons

image
image


packages/ui/__stories__/TextField.stories.module.scss
```diff
@@ -1,3 +1,10 @@
+@use '../styles/constants' as c;
+
 .field {
   margin-bottom: 16px;
 }
+
+.shortField {
+  width: c.$grid * 25;
```
Sodik, line 8: I think it should be done on parent element in app not in hive. What if I have to use it with bigger numbers (billions e.t.c)?

Sodik (changes requested):

packages/ui/__stories__/NumberField.stories.tsx
```diff
@@ -33,7 +33,9 @@ const Template: Story<NumberFieldProps> = ({ value: initValue, ...args }) => {
   return (
     <>
       <NumberField {...args} value={value} onChange={setValue} />
+      <NumberField {...args} value={value} onChange={setValue} iconPosition="together" className={storyStyles.shortField} />
```
fantkolja, line 36: @allfayn @pawelklasa Should we do something with placeholder?
image

fantkolja (changes requested):

packages/ui/__stories__/TextField.stories.module.scss
allfayn: It's not in Hive, it's only for presentation (at Story)

fantkolja (approved):

Sodik (approved):

packages/ui/__stories__/NumberField.stories.tsx
allfayn: removed placeholders for this kind of fields at story

PR #422, opened 2021-12-20
UI 241 tree shaking (HZC-2157)

### LINKED ISSUE #241: Tree shaking
Seems like the libraries are not tree-shakable (checked on Cloud, but only once and not 100% sure).
Need to investigate and if is really not - implement it

fantkolja:
Closes #241 

Here is the result on cloud UI
image

Note: this is a breaking change since I haven't included type exports into our `index.ts` export files. However, you can always import types from the nested file.


utkukaratas:
@fantkolja awesome result! one question - what was the reason you got rid of all the star* exports?

utkukaratas (approved):

fantkolja:
> @fantkolja awesome result! one question - what was the reason you got rid of all the star* exports?

Actually, as I understood from many articles I have read, those kinds of exports prevent tree shaking, since the bundler can't identify, which exactly entities are used and which aren't. 
As you can see replacing `*` exports with the common named exports is basically everything I did here to achieve the result.

allfayn (approved):

PR #442, opened 2022-01-20
feat(ui): custom name for tabs

allfayn:
image

image


packages/ui/src/Tabs/Tab.tsx
```diff
@@ -19,22 +19,34 @@ export type ButtonTabProps = {
...
+  | {
+      label?: never
+      icon?: never
+      iconAriaLabel?: never
+      children: ReactNode
+      ariaLabel: string
+    }
+
 export type TabCommonProps = {
-  label: string
```
fantkolja, line 32: How about we just make the type of label `ReactNode` instead of creating a third set of TabLabelProps?
WDUT @utkukaratas @Sodik?

fantkolja: And by the way - this is our global problem. Almost all of our`label` props are `string` but it's very often the case, that you need to pass something "bigger" than a string, for example a string with a `span` or a string with an Icon.

fantkolja: I would say we could even make a separate PR, where we upgrade all of our labels from a `string` to the `ReactNode`

allfayn: I think, that using `label` as `ReactNode` is a not good practice at all. `Label` is a text, for `ReactNode` we have a `children`. 
When we use `label` as `ReactNode`, we can have some problems with other properties, like Icons.
In this case, using `ReactNode` is opposite to using `label` as text + `icon`.

allfayn: In my opinion this:
```
<tab value={0}>Tab <b>1</b></tab>
```
better than that:
```
<tab value={0} label={Tab <b>1</b>}/>
```

allfayn: And I want to point out that this is not just a code problem. This is an important feature in using the library and its root API.

allfayn: In case when we use `label` as `ReactNode` we can do that:
```<tab value={0} label={Tab <b>1</b>} icon={Info} iconAriaLabel="Info" />```
The result will be not predictable at all, I guess. And typing will not help as in this situation.

fantkolja: 1. If the client wants to put `Icon` into `icon` and then the same `Icon` as a part of the label it's his problem.
2. I think it's pretty predictable - there will be 2 `Icon`s :)

allfayn: And the link with a discussion like this 
https://stackoverflow.com/questions/56003891/react-props-vs-children-what-to-use-when

fantkolja (approved):

utkukaratas (approved):

PR #449, opened 2022-02-08
fix(ui): remove overflow for self portal target in Select Component

allfayn:
Changes:
- remove `overflow: hidden` in Select Component for using `menuPortalTarget="self"`
- add and throw prop `closeMenuOnScroll` from `react-select` library


utkukaratas (approved):
is loki ok? there's are too many screenshot changes.

fantkolja (approved):

packages/ui/src/Select/SelectField.tsx
```diff
@@ -56,7 +56,10 @@ export type SelectFieldExtraProps<V> = {
   singleValueTooltipVisible?: boolean
 } & DataTestProp &
   Pick<InputHTMLAttributes<HTMLElement>, 'autoFocus' | 'disabled' | 'required' | 'placeholder'> &
-  Pick<ReactSelectProps, 'isSearchable' | 'isClearable' | 'menuIsOpen' | 'menuPlacement' | 'noOptionsMessage' | 'inputValue'> &
+  Pick<
+    ReactSelectProps,
+    'isSearchable' | 'isClearable' | 'menuIsOpen' | 'menuPlacement' | 'closeMenuOnScroll' | 'noOptionsMessage' | 'inputValue'
+  > &
```
Sodik, line 62: Please enable closeMenuOnScroll when menuPortalTarget = 'body'

Sodik (changes requested):

packages/ui/src/Select/SelectField.tsx
allfayn: It's experimental.
I have a plan to do this later, after checking how it works in our application.
This option doesn't seem to work properly with complex pages and iframe (storybook). Maybe we need a new code to do this. 
And changes in loki's screenshots are very small and minor

Sodik (approved):

PR #459, opened 2022-04-21
Close dropdown menu in SelectField while resizing

aberatcan:
- Dropdown menu is misplaced in SelectField while resizing. 
- It is stem from the react-select library issue
- For more information visit https://github.com/JedWatson/react-select/issues/3533

### DIFF
--- a/packages/ui/src/Select/MultiSelectField.tsx
+++ b/packages/ui/src/Select/MultiSelectField.tsx
@@ -165,6 +165,22 @@ export const MultiSelectField = <V extends string | number = number>(props: Mult
     }
   }, [menuIsOpen])
 
+  /** If the size of screen is changed we close dropdown menu to aviod the dropdown menu bug in react-select library.
+   * For more information visit: https://github.com/JedWatson/react-select/issues/3533
+   */
+  useEffect(() => {
+    const handleResize = () => {
+      if (selectRef.current) {
+        selectRef.current.blur()
+      }
+    }
+
+    window.addEventListener('resize', handleResize)
+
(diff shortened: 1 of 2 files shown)


packages/ui/src/Select/MultiSelectField.tsx
```diff
@@ -165,6 +165,22 @@ export const MultiSelectField = <V extends string | number = number>(props: Mult
     }
   }, [menuIsOpen])
 
+  /** If the size of screen is changed we close dropdown menu to aviod the dropdown menu bug in react-select library.
```
Sodik, line 168: Cause we use the same functionality in Select it is better to move it into a custom hook.

packages/ui/src/Select/MultiSelectField.tsx
```diff
@@ -165,6 +165,22 @@ export const MultiSelectField = <V extends string | number = number>(props: Mult
...
+   * For more information visit: https://github.com/JedWatson/react-select/issues/3533
+   */
+  useEffect(() => {
+    const handleResize = () => {
+      if (selectRef.current) {
+        selectRef.current.blur()
```
Sodik, line 174: Doesn't `selectRef.current.onMenuClose()` work?

Sodik (changes requested):

packages/ui/src/Select/MultiSelectField.tsx
aberatcan: > Doesn't `selectRef.current.onMenuClose()` work?

It also works but need ReactSelect reference type.

Sodik (approved):

PR #467, opened 2022-05-26
Replace npm to pnpm

allfayn:
- replaced npm with pnpm (use for workspaces)
- updated Lerna v6 (use for publishing only)
- updated React v17
- replaced Jest transformer to Babel
- updated other dependencies
- fixes and improvements


README.md
```diff
@@ -91,19 +91,14 @@ Uses stories from [Storybook](https://storybook.js.org/) as test cases. In other
 ### Setup
```
Sodik, line 91: Maybe we should describe how to install required pnpm version. I'm not sure we could easily migrate to next major version.
Ex. `npm i -g pnpm@8.3.1`

Sodik (approved):

utkukaratas (approved):

fantkolja (approved):

PR #471, opened 2022-06-29
update(ui): can use react elements in Button component

allfayn:


Sodik (approved):

packages/ui/src/Button.tsx
```diff
@@ -237,7 +235,7 @@ export const Button = forwardRef<HTMLElement, ButtonProps>(
                 />
               )}
               <TruncatedText
-                text={capitalize ? children.toUpperCase() : children}
+                text={capitalize && typeof children === 'string' ? children.toUpperCase() : children}
```
fantkolja, line 238: if we pass ReactElement instead of a simple text we shouldn't use `TruncatedText` I guess...

fantkolja (changes requested):

packages/ui/src/Button.tsx
allfayn: It is used to add tooltip to text/element here. And it's not a text clipping function at all.

PR #472, opened 2022-07-12
update(code): add simple default format date/time function

allfayn:


fantkolja (approved):

packages/services/src/time.ts
```diff
@@ -3,4 +3,20 @@ export class TimeService {
...
+    return `${TimeService.formatDate(source)} ${TimeService.formatTime(source)}`
+  }
+
+  public static formatTime(source: Date): string {
+    return `${('0' + source.getHours().toString()).slice(-2)}:${('0' + source.getMinutes().toString()).slice(-2)}:${(
+      '0' + source.getSeconds().toString()
+    ).slice(-2)}`
+  }
+
+  public static formatDate(source: Date): string {
```
Sodik, line 17: Why do you need it?
We have to use  "YYYY-MM-DD" instead.
This format was approved across all products

Sodik (changes requested):

packages/services/src/time.ts
allfayn: It is the same formatting without the involvement of third-party libraries and regardless of client localization.
`YYYY-MM-DD hh:mm:ss`

What function/library do you use for format date string?

Sodik: formatDate method returns 'DD-MM-YYYY' instead 'YYYY-MM-DD'
We use moment

allfayn: Thanks. Fixed format to `YYYY-MM-DD hh:mm:ss`

Sodik (approved):

PR #475, opened 2022-08-17
*fix Clear unselected table cell values

Sodik:
*fix Ignore hotkeys on Table sub rows

### DIFF
--- a/packages/ui/src/Table/Table.tsx
+++ b/packages/ui/src/Table/Table.tsx
@@ -214,7 +214,7 @@ type CustomTableProps<D extends object> = {
   columnsOrdering?: boolean
   children?: (table: ReactElement, tableInstance: TableInstance<D>) => ReactElement
   renderRowSubComponent?: (props: RowType<D>) => ReactNode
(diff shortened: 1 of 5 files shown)


fantkolja (approved):

packages/ui/src/Table/Table.tsx
```diff
@@ -214,7 +214,7 @@ type CustomTableProps<D extends object> = {
   columnsOrdering?: boolean
   children?: (table: ReactElement, tableInstance: TableInstance<D>) => ReactElement
   renderRowSubComponent?: (props: RowType<D>) => ReactNode
-  onCopy?: (data: string[][]) => void
+  onCopy?: (data: (string | undefined)[][]) => void
```
utkukaratas, line 217: how about "data?: ..."

utkukaratas (approved):

packages/ui/src/Table/Table.tsx
Sodik: Data can be - `[undefined, undefined, 'string']`

PR #476, opened 2022-09-02
interactive list disabled state

allfayn:
Disables state for Interactive List Component

image

### DIFF
--- a/packages/ui/src/InteractiveList.tsx
+++ b/packages/ui/src/InteractiveList.tsx
@@ -10,6 +10,8 @@ import { FieldHeaderProps } from './FieldHeader'
 
 export type InteractiveListExtraProps = {
   children?: React.ReactNode | React.ReactNode[]
+  disabled?: boolean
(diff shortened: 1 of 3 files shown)


packages/ui/src/InteractiveList.tsx
```diff
@@ -112,8 +120,10 @@ export const InteractiveList = <V,>({
           kind="transparent"
           ariaLabel="Add Icon"
           icon={Plus}
-          className={styles.addIcon}
+          className={cn(styles.addIcon, { [styles.disabledIcon]: disabled })}
           size="medium"
+          disabled={!!disabled}
```
Sodik, line 125: `disabled={disabled}`

Sodik (approved):

packages/ui/src/InteractiveList.tsx
allfayn: I know, but it is for a type check. `disabled: boolean | undefined`

PR #493, opened 2023-05-04
better accessibility score with ARIA updates

aberatcan:
- updated aria-label of `SelectField`
- given an id to each `LinkRow` and added `aria-owns` label
- To increase to Accessibility Score of Lighthouse Tests

### DIFF
--- a/packages/ui/src/Select/SelectField.tsx
+++ b/packages/ui/src/Select/SelectField.tsx
@@ -154,7 +154,7 @@ export const SelectField = <V extends string | number = string>(props: SelectFie
     inputId: id,
     className: 'hz-select-field',
     classNamePrefix: 'hz-select-field',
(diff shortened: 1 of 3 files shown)


packages/ui/src/Select/SelectField.tsx
```diff
@@ -154,7 +154,7 @@ export const SelectField = <V extends string | number = string>(props: SelectFie
     inputId: id,
     className: 'hz-select-field',
     classNamePrefix: 'hz-select-field',
-    'aria-label': showAriaLabel ? label : undefined,
+    'aria-label': showAriaLabel ? label : 'Select Field',
```
utkukaratas, line 157: "select" would be better. "field" i dunno.

utkukaratas (approved):

PR #494, opened 2023-05-18
feat(tooltip): change default word breaking at help texts

allfayn:


fantkolja (approved):

packages/ui/src/Tooltip.module.scss
```diff
@@ -19,7 +19,7 @@ $tooltipArrowSize2-3: math.div(c.$tooltipArrowSize, 3) * 2;
   border: c.$borderWidth solid c.$colorNeutralLight;
   border-radius: c.$borderRadius;
   color: c.$colorText;
-  word-break: break-all;
+  word-break: normal;
```
Sodik, line 22: It will break some of MC tables.
Please make it configurable

Sodik (changes requested):

PR #498, opened 2023-06-21
*fix Improve table columns resizing

Sodik:
*feat Table context menu


packages/ui/src/Table/features/columnsSelection/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useMemo } from 'react'
```
allfayn, line 1: Should we add e2e tests for the new component?

Sodik: We should. I just don't know when we could setup it

allfayn: I talking about screenshot/loki tests, that we have already.

Sodik: done

utkukaratas (approved):

PR #513, opened 2023-09-04
MC-2117 Watch column size and page size changes

emrberk:
Management Center has to keep track of column resizing and page size selection. That's why, the changes in `pageSize` and `columnResizing` will be passed to the callback, obtained from MC.
Also, to prevent exhaustive calls to the passed callback, I added `useDebouncedEffect`, which calls the effect function once in minimum of 200ms. The delay is configurable.

### DIFF
--- a/packages/ui/src/Table/Table.tsx
+++ b/packages/ui/src/Table/Table.tsx
@@ -29,6 +29,7 @@ import {
   useColumnOrder,
   useExpanded,
   TableInstance,
+  TableState,
 } from 'react-table'
 import { AlertTriangle, ChevronDown, ChevronUp } from 'react-feather'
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/Table/Table.tsx
```diff
@@ -422,6 +426,12 @@ export const Table = <D extends object>({
     [onKeyUp],
   )
 
+  useDebouncedEffect(() => {
```
Sodik, line 429: Why don't you use this option?
`children(content, tableInstance)`

emrberk: Is `children` supposed to be a function component, which will listen to changes by passed `tableInstance` argument, and will return `content`? If this is the case, the call `children(content, tableInstance)` is not suitable (as it does not have children argument).
If it does not work like that, we can't utilize hooks inside `children` function. We will have to call the hook inside this file (to achieve the same effect), which will prevent us to get rid of the changes in Hive.

emrberk:
This PR is closed for the sake of https://github.com/hazelcast/management-center/pull/7929

PR #524, opened 2023-11-07
Remove focus trap from Popover and disable Table context menu on header cells

Sodik:
*fix remove focus trap from popover
*fix Table context menu appearance on header cells

### DIFF
--- a/packages/ui/src/Popover.tsx
+++ b/packages/ui/src/Popover.tsx
@@ -43,7 +43,7 @@ export const Popover: FC<PopoverProps> = (props) => {
   } = props
 
   const getOnUpdate = useRefValue(onUpdateLayout)
-  const getOnClos = useRefValue(onClose)
+  const getOnClose = useRefValue(onClose)
   const [popperElement, setPopperElement] = useState<HTMLElement | null>(null)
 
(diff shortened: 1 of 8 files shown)


packages/ui/src/Popover.tsx
```diff
@@ -42,16 +41,11 @@ export const Popover: FC<PopoverProps> = (props) => {
     onUpdateLayout,
     container = `body`,
   } = props
+
   const getOnUpdate = useRefValue(onUpdateLayout)
+  const getOnClos = useRefValue(onClose)
```
aberatcan, line 46: Typo here `getOnClose` ?

aberatcan (approved):

packages/ui/src/Popover.tsx
Sodik: Thanks

fantkolja (approved):

packages/ui/src/Popover.tsx
```diff
@@ -111,32 +105,53 @@ export const Popover: FC<PopoverProps> = (props) => {
     handler: onClose,
   })
 
+  const handleFocusIn = useCallback(
+    (e: FocusEvent) => {
+      console.log(e.target)
```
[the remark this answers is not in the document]
Sodik: done

PR #529, opened 2023-11-14
fix(ui): update focus styles at Tab and use focus-visible

### LINKED ISSUE #319: Accessibility outline fix
A good example of this is a bbc.co.uk website where the outline is hidden upon clicking but visible on TAB

allfayn:
use `:focus-visible`

image

fixed #319


Sodik (approved):

Sodik:
@allfayn Focus style does not aligned with other component. Why do you need it only for one component?
There was an idea to use custom focus only on keyboard focus for all components not for only one

Sodik:
image
image

PR #535, opened 2024-01-18
feat(ui): add close button to Notification

emrberk:
Management Center has Healthcheck banner using Notification component, indicating the problem count. The user should be able to close the banner, especially when there's no action to be taken.
This PR adds a close button to Notification component, and leaves the close logic to the parent component via callback.
Screenshots are updated.


fantkolja (approved):

packages/ui/src/Notification.tsx
```diff
@@ -84,6 +86,7 @@ export const Notification: FC<NotificationProps> = ({ type, children, className,
           {link}
         </Link>
       )}
+      {onClose && <IconButton className={styles.closeButton} icon={X} onClick={onClose} ariaLabel="Notification close button" />}
```
Sodik, line 89: What about unit test?

Sodik (changes requested):

packages/ui/src/Notification.tsx
emrberk: Added 👍

Sodik (approved):

PR #543, opened 2024-04-18
Avoid overflow of a long text value inside SelectField in Firefox [MC-2784]

emrberk:
After upgrading to react-select to v5.8.0 from v4.3.1, select field values with long text started to overflow from the container. Restricted the maximum width to 100%.
There is a related [issue](https://github.com/JedWatson/react-select/issues/5170), but no one seems interested :)


fantkolja (approved):

emrberk:
But this one breaks the trailing note. Recalculating the route...

Sodik (approved):

PR #578, opened 2024-11-07
new card variant

Sodik:


emrberk:
Why do we omit paddings in the card?

Sodik:
> Why do we omit paddings in the card?

not needed anymore
@pawelklasa

pawelklasa (approved):
This new card variant is required for MC dashboard modules

PR #579, opened 2024-11-19
*feat basic Table redesign

Sodik:


packages/ui/src/Table/Row.module.scss
```diff
@@ -18,18 +13,17 @@
       // Do not remove outline for high-contrast modes
       // https://stackoverflow.com/a/52616313
       outline: c.$outlineWidth solid transparent;
-      background-color: c.$colorInfoLight;
+      background-color: #e2e7fa;
```
fantkolja, line 16: Maybe at least a local variable, since it's used several times

packages/ui/styles/typography/ubuntu-mono.css
```diff
@@ -0,0 +1,31 @@
...
+  font-family: 'Ubuntu Mono';
+  font-style: normal;
+  font-display: swap;
+  font-weight: 400;
+  src: url('./fonts/UbuntuMono-Regular.ttf') format('ttf');
+}
+
+@font-face {
+  font-family: 'Ubuntu Mono';
+  font-style: italic;
```
fantkolja, line 11: Do we need italic monospace font?

packages/ui/styles/typography/ubuntu-mono.css
```diff
@@ -0,0 +1,31 @@
+@font-face {
+  font-family: 'Ubuntu Mono';
```
fantkolja, line 2: https://hazelcast.slack.com/archives/C01BT11F24E/p1732781421xxxx11

packages/ui/styles/typography/ubuntu-mono.css
```diff
@@ -0,0 +1,31 @@
...
+  font-style: normal;
+  font-display: swap;
+  font-weight: 400;
+  src: url('./fonts/UbuntuMono-Regular.ttf') format('ttf');
```
fantkolja, line 6: I would only include woff2
https://stackoverflow.com/questions/11002820/why-should-we-include-ttf-eot-woff-svg-in-a-font-face

fantkolja (changes requested):

packages/ui/styles/typography/ubuntu-mono.css
Sodik: Custom content may need it

packages/ui/src/Table/Row.module.scss
Sodik: Colors are not finalized yet. There is a separate task for that.

fantkolja (approved):

PR #634, opened 2026-05-13
HZC-7919 Add fullWidth prop to Button

pawelklasa:
## Summary

Adds a new `fullWidth` boolean prop to `Button` so it can stretch to fill its container, overriding the default `inline-flex` layout and the `--hive-button-max-width` (370px) cap.

## Changes

- `src/components/Button.tsx` — new `fullWidth?: boolean` prop on `ButtonCommonProps`; applies `styles.fullWidth` conditionally.
- `src/components/Button.module.css` — `.fullWidth { display: flex; width: 100%; max-width: 100%; }`.
- `__stories__/Button.stories.tsx` — `FullWidth` story now uses the prop instead of a local block class.
- `__tests__/components/Button.test.tsx` — coverage for both \`fullWidth\` true and default.

## Tests

- \`npm test -- --testPathPattern=Button\` ✅ (40 tests)
- \`npm run lint\` ✅

## Jira

HZC-7919

### DIFF
--- a/src/components/Button.module.css
+++ b/src/components/Button.module.css
@@ -47,6 +47,12 @@
   --button-padding-x: 16px;
 }
 
+.fullWidth {
+  display: flex;
+  width: 100%;
+  max-width: 100%;
+}
+
 .button:hover,
 .button.hover {
   background: var(--button-bg-hover);
--- a/src/components/Button.module.css
+++ b/src/components/Button.module.css
@@ -47,7 +47,7 @@
   --button-padding-x: 16px;
 }
 
-.fullWidth {
+.fullWidth.fullWidth {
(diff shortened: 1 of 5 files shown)


src/components/Button.module.css
```diff
@@ -47,6 +47,12 @@
   --button-padding-x: 16px;
 }
 
+.fullWidth {
+  display: flex;
+  width: 100%;
+  max-width: 100%;
```
fantkolja, line 53: That might not work in some scenarios because those styles have the same specificity as the base styles

fantkolja (changes requested):

fantkolja (approved):

PR #633, opened 2026-05-07
hzc-7906: Restyle SelectField and AutocompleteField to HIVE v4

pawelklasa:
## HZC-7906

Restyle the entire Select stack and AutocompleteField to HIVE 4.0.

### Scope
- SelectField, MultiSelectField, CheckableSelectField (+ Formik wrappers, Common, helpers)
- AutocompleteField (+ Formik wrapper)

### Changes
- v4 CSS modules with --hive-color-*-v4 tokens, --hive-grid, --hive-border-radius, --hive-border-width
- 36px control height, 8px radius, hover -> text-v4 border, focus -> primary-v4 + 3px halo (matches TextField/Button)
- Menu: 8px radius, soft shadow, 40px option rows, hover surface-hover-v4, selected surface-active-v4 + medium weight
- Open-state control drops the focus halo so it looks attached to the menu
- v3 fallback under src/old, exported as @hazelcast/ui/old
- Storybook: Playground + MDX docs (Overview / Guidelines / Accessibility) + LegacyV3 story for each component
- Migration notes in docs/migration-v4.md

### Tests
- 47/47 Select + Autocomplete tests pass
- Lint clean

PR #637, opened 2026-06-03
hzc-8000: redesign the SelectField component HIVE v4

pawelklasa:
## Summary
Master's HIVE v4 SelectField (from #633 / hzc-7906) renders the dropdown chevron via `Icon size="medium"` = **24px**, but the Figma spec calls for **16px**. This adds an explicit rule sizing the indicator svg down.

## Change
```css
.container :global(.hz-select-field__indicators) svg {
  width: calc(var(--hive-grid) * 4);  /* 0.25rem * 4 = 16px */
  height: calc(var(--hive-grid) * 4);
}
```

## Why it works
The svg's `width="1em"` is a presentational *attribute* (lowest cascade priority), so this CSS rule wins over `Icon`'s `.medium` rule on specificity — no `!important` needed.

## Verification
Confirmed live in Storybook (`components-selectfield--basic`): the chevron renders at exactly **16×16px** (bounding box + computed style).

## Scope
Self-contained follow-up to the merged hzc-7906 redesign. Touches only `SelectField.module.css`.

### DIFF
--- a/src/components/Select/SelectField.module.css
+++ b/src/components/Select/SelectField.module.css
@@ -144,7 +144,7 @@
 }
 
 .menuContainer.menuContainer.menuContainer :global(.hz-select-field__menu-list) {
-  padding: calc(var(--hive-grid) * 1) 0;
+  padding: 0;
   color: var(--hive-color-text-v4);
 }
 
--- a/src/components/Select/SelectField.module.css
+++ b/src/components/Select/SelectField.module.css
@@ -50,6 +50,11 @@
   display: none;
 }
 
+.container :global(.hz-select-field__indicators) svg {
+  width: calc(var(--hive-grid) * 4);
+  height: calc(var(--hive-grid) * 4);
+}
+
 .container :global(.hz-select-field__input-container),
 .container :global(.hz-select-field__input) {
   color: var(--hive-color-text-v4);
--- a/src/components/Select/SelectField.module.css
+++ b/src/components/Select/SelectField.module.css
@@ -50,11 +50,6 @@
   display: none;
 }
 
-.container :global(.hz-select-field__indicators) svg {
-  width: calc(var(--hive-grid) * 4);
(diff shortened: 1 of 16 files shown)


src/components/Select/SelectField.module.css
```diff
@@ -50,6 +50,11 @@
   display: none;
 }
 
+.container :global(.hz-select-field__indicators) svg {
```
fantkolja: Probably not related but I guess this padding is unwanted 

Image

src/components/Select/SelectField.module.css
```diff
@@ -50,6 +50,11 @@
...
+  width: calc(var(--hive-grid) * 4);
```
fantkolja: If you set size small and isClearable true the size goes to medium 

Image

fantkolja: Medium size also increases the label font 

Image

src/components/Select/SelectField.module.css
```diff
@@ -50,6 +50,11 @@
...
+  height: calc(var(--hive-grid) * 4);
```
fantkolja: I don't get what these properties do 

Image

fantkolja (changes requested):

src/components/Select/SelectField.module.css
pawelklasa: Fixed

src/components/Select/SelectField.module.css
pawelklasa: Fixed

pawelklasa: Fixed

src/components/Select/SelectField.module.css
pawelklasa: fixed

fantkolja (approved):

